PR #45, opened 2014-05-23
Created simple functional pattern matching construct.

dkrieg:
Please consider this contribution.


mperry:
I created some similar, but better classes to implement the stream processing chapter from the `FP in Scala` book, see https://github.com/mperry/stream-processing/tree/master/src/main/java/com/github/mperry/fj.  This is an improvement and was useful to implement the chapter, but not sure if it is useful enough to include.  I will close this PR for now and we can reconsider this if people have opinions they'd like to share.

PR #68, opened 2014-12-24
Add Lenses - Getter and Setter for immutable data structures

knutwalker:
A Lens is the purely functional alternative for getters and setters.
The implementation is bases one the one provided by scalaz, which
is based on Kmett's Haskell Lenses.

### DIFF
--- a/core/src/main/java/fj/data/Lens.java
+++ b/core/src/main/java/fj/data/Lens.java
@@ -1,0 +1,512 @@
+package fj.data;
+
+import fj.F;
+import fj.F2;
+import fj.Function;
+import fj.P;
+import fj.P1;
+import fj.P2;
+import fj.Unit;
+
+/**
+ * A Lens offers a purely function means to access and retrieve a field in a
+ * record.
+ *
+ * The term {@code field} does not necessarily mean an actual field or member
+ * of a class. For example, a lens can address membership of a {@code Set}.
+ *
+ * @param <A> The type of the record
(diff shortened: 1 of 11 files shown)


mperry:
@tonymorris I am not familiar with twanvl lenses, but is this a useful addition given Java's many limitations?  I would imagine a twanvl lens representation would be a lot of extra work if it needed kinds immediately.  

If we changed to twanvl lenses later, how would clients of a lens be affected?

jbgi:
May be the avatar of higher order types as implemented in https://code.google.com/p/highj/ could be a solution? But the verbose generics are a bit repulsive...

danieldietrich:
To get in the direction of higher order types I personally take a naive approach based on generics: https://gist.github.com/danieldietrich/6320d0d617ededbe05f7

I'm using it here for example: http://javaslang.com/javadoc/javaslang/monad/Option.html#flatMap-java.util.function.Function-

jbgi:
@danieldietrich you solution is technically very similar to the one used in highj. Except highj has better separation of concerns, ie List is not required to extends all the "pseudo type-classes" interfaces (foldables and the like). This allows creation of custom "type-classes" in client code.

danieldietrich:
@jbgi thank you for the feedback. I will take a look at highj!

mperry:
I am going to close this PR, since this is addressed by @jbgi in #130 #131 and #132.

PR #69, opened 2014-12-24
Add some Collectors for Java8

knutwalker:


mperry:
@knutwalker Thanks for your contribution.

PR #71, opened 2015-01-04
Add toInt and fromInt to fj.Ordering.

dobesv:
I thought it was handy to be able to convert from the standard Ordering type in Java - the int - and the fj Ordering more easily.


techtangents:
Looks good to me.

PR #72, opened 2015-01-04
Add MANIFEST.MF so this project can be used as an eclipse plugin.

dobesv:
I've been using functionaljava in my eclipse projects, this change might be handy for others as well.


META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,21 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Functional Java
+Bundle-SymbolicName: org.functionaljava
+Bundle-Version: 4.3.0.qualifier
```
mperry: @dobesv I don't know about Eclipse's use of manifests, but should the bundle version above be 4.2 for the last release?

dobesv: Hmm it's a good point
Also I have since learned that gradle is generating the manifest
currently.  I should redo this one.

mperry:
@dobesv Ok, let us know when you have reviewed and updated the PR.

mperry:
You might have figured this out, but you can set the attributes in the manifest by editing the core/build.gradle file and adding this:

```jar {
    manifest {
        attributes("attribute1": "attribute1 value", "attribute2": "attribute2 value")
    }
}
```

dobesv:
Yes, I did see that.  But on further reflection I realized that if I am
working with the project in eclipse I actually want the MANIFEST.MF to be a
static file so that eclipse can use it.

I have convinced gradle to build an OSGI jar file and also output a
MANIFEST.MF for eclipse to use.  I'll sit on it a little longer, though, to
think about possible repurcussions.  Do you think it's a problem if
functionaljava is always built as an OSGI jar?  AFAIK you can use them as
normal jar files.

mperry:
I haven't used Eclipse for more than 3 years, can you explain why you want to use projects as plugins in Eclipse?  A quick Google search wasn't helpful.

dobesv:
I am using it in an eclipse plugin, so converting the jar to an eclipse
plugin seemed to make it easier to get on the class path of my plugin.

I can probably maintain this as a little fork of my own indefinitely.  I'm
not sure it will ever be of use to anyone else.

mperry:
I spent 10 minutes reading around and it looks like the extra info in the manifest required for an OSGI bundle is ignored in most applications.  I think generating the info from the Gradle build is feasible.

dobesv:
Yes, I have this working locally now, I think it will work fine in and out
of eclipse.

One hiccup I ran into is that the OSGI plugin for gradle doesn't like the
tests project because it doesn't create the folder to contain the main
class files. Do we actually need to build and distribute jar files for the
tests?  I added a workaround to create the missing folder first, which
works, but I wonder if I could disable building jar files for the tests
altogether.

mperry:
The tests project is separate because of the various different support for Scala in IDEs.  Putting them in a separate Gradle module allows us to exclude that module temporarily (if necessary).  

We don't need to build the jar, just to run the tests.

mperry:
So where do we go from here?

dobesv:
Hi Mark,

Sorry I have been a bit slow on my PR updates, I've just been a bit busy
work other things.  I think I will make the changes to generate an OSGI
compatible jar for the projects except tests, plus generate a MANIFEST.MF
if you run the eclipse target, and update the PR with those changes for
review.

dobesv:
OK so I updated it so that it builds an OSGI jar for each subproject except tests.  Hopefully this doesn't interfere with anyone else's needs.  Take a look and tell me what you think.

mperry:
FYI, I probably won't get to looking at this until the weekend.

mperry:
I think this is ok, everyone else will just see a more detailed manifest in their jar where the keys will be ignored.

PR #80, opened 2015-01-12
Adding traverse to List, Option, Either and Tuple2

amarpotghan:


tonymorris:
FWIW, this is specifically `List#traverse`. You might also consider writing `traverse` for every possible `Traversable`. Unfortunately, that is a lot of work, since there are infinity of them, so as a compromise, perhaps the more common ones; `List`, `Option`, `Either<T, _>`, Tuple2<A, _>

amarpotghan:
Oops...i totally did not see that. I agree that it's a list traverse.  
I was planning to wtite it for every traversable, but as u said its a lot of work. That's why decided this specific one...because i found its handy

tonymorris:
Yeah no worries. You'll probably want the others some day too!

amarpotghan:
@tonymorris sure thing :+1:

mperry:
@tonymorris and @amarpotghan 
For some reason I thought traverse was defined on Monad with the signature:
`traverse:: Monad f => List a -> (a -> m b ) -> m (t b)`
whereas it is actually on Traversable:
`traverse :: Applicative f => (a -> f b) -> t a -> f (t b)`

Given the use of different applicatives, is the best way to handle this to represent the function on a traversable type t as:
`public <B> ${Applicative}<${Traversable}<B>> traverse${Applicative}(F<A, ${Applicative}<B>>)`
so for this example by @amarpotghan for traversing lists we would have on class List:
`public <B> Option<List<B>> traverseOption(F<A, Option<B>)`

Some other classes this would apply to that I use: IO(?), SafeIO, Seq, Set(?), Stream, P1, Validation<A, _>, State<S, _>.

This of course is starting to create alot of combinations, but it would be good to agree on the approach and fill in the missing implementations at some future point.

tonymorris:
Yep that's the problem. You have NxM instances to write where:

* N = Number of `Applicative` instances
* M = Number of `Traversable` instances

Given the limitations of Java's type system against the intractability of this problem, Functional Java has had a history of, "just pick the most common ones that I have the energy to write today."

tonymorris:
Consequently, I cannot speak for the energy of @amarpotghan.

amarpotghan:
@mperry No doubt I have copied the documentation of `sequence` without bothering to read it :(
I like the suggested representation. 

The only problem I see as @tonymorris mentioned - the number of combinations required. I have the energy to write a few of them ;), so maybe let's choose the most handy ones. And let's put others on to-do radar. 

That reminds me about asking - where do we manage to-do things?

mperry:
For TODO things, we should create Github issues.  We can decide what we want to do for traverse here and then create an issue documenting what is left to do.

I had a look at doing this for the Option type last night (so don't worry about Option).  BTW, I also had a look at the laws that traversal had to comply with, but only really understood the identity law.  I suspect this won't be a barrier to writing the code.

amarpotghan:
@mperry There is also an traverseIO in the IOFunctions. So we might not need to write it again. Also mapMOption is implemented in the List type(I did not see that earlier). I will refactor it to express it in terms of traverse, what do you think?

amarpotghan:
@mperry I think this can be merged now. Could you please have a look and suggest changes if any?

core/src/main/java/fj/P1.java
```diff
@@ -167,11 +132,57 @@ public B _1() {
...
+    /**
+     * Traversable instance of P1 for Option
+     *
+     * @param f The function might successfully might fail
+     * @return An Option of  P1<B>
+     */
+    public <B> List<P1<B>> traverseOption(final F<A, List<B>>  f){
+        return f.f(_1()).map(b -> P.p(b));
+    }
+
```
mperry: I think you have copy and pasted and forgotten to change the signature to include option here.

core/src/main/java/fj/P1.java
```diff
@@ -167,11 +132,57 @@ public B _1() {
...
+    /**
+     * Traversable instance of P1 for Either
+     *
+     * @param f The function might successfully return B or might fail with an error notification of type X
+     * @return An Stream of  P1<B>
+     */
+    public <B> List<P1<B>> traverseStream(final F<A, List<B>>  f){
+        return f.f(_1()).map(b -> P.p(b));
     }
```
mperry: Looks like you have forgotten to update this for a stream.

core/src/main/java/fj/P1.java
```diff
@@ -237,10 +240,11 @@ public String toString() {
 		return Show.p1Show(Show.<A>anyShow()).showS(this);
 	}
 
+    @SuppressWarnings("unchecked")
     @Override
     public boolean equals(Object other) {
-        return !Equal.equalsValidationCheck(this, other) ? false :
-                Equal.p1Equal(Equal.<A>anyEqual()).eq(this, (P1<A>) other);
+        return Equal.equalsValidationCheck(this, other)
+                && Equal.p1Equal(Equal.<A>anyEqual()).eq(this, (P1<A>) other);
```
mperry: Thanks for point this out.  In the future, please do this as part of a different change request.

core/src/main/java/fj/P1.java
amarpotghan: oops..

mperry:
No problem.  FYI, I started reviewing these changes, but have not had time to finish yet.  I will try to do some more tonight.

amarpotghan:
@mperry thanks

core/src/main/java/fj/P1.java
amarpotghan: sure @mperry . I thought that is too minor to have a separate PR. Affirmative to the point of tracking things separately.

core/src/main/java/fj/data/Either.java
```diff
@@ -509,6 +530,41 @@ public void foreachDoEffect(final Effect1<B> f) {
...
+       *
+       * @param f the function to traverse with
+       * @return An either after traversing through this projection.
+    */
+      public <C> List<Either<A, C>> traverseList(final F<B, List<C>> f) {
+          return isRight() ?
+                  f.f(value()).map(x -> Either.right(x)) :
+                  list(Either.<A, C>left(e.left().value()));
+      }
+
```
mperry: I commented above that you were traversing over the left value and here you are traversing over the right value.  I did not think you would do both.  Is there a conflict here for the compiler in calls to traverseList(z -> ...)?  Should this be renamed to traverseLeftList/traverseRightList?

core/src/main/java/fj/data/Either.java
```diff
@@ -724,11 +752,7 @@ public boolean exists(final F<B, Boolean> f) {
   public static <A, X> Either<List<A>, X> sequenceLeft(final List<Either<A, X>> a) {
     return a.isEmpty() ?
            Either.<List<A>, X>left(List.<A>nil()) :
-           a.head().left().bind(new F<A, Either<List<A>, X>>() {
-             public Either<List<A>, X> f(final A aa) {
-               return sequenceLeft(a.tail()).left().map(cons_(aa));
-             }
-           });
+           a.head().left().bind(aa -> sequenceLeft(a.tail()).left().map(cons_(aa)));
```
mperry: Note the separate calls to sequenceLeft and sequenceRight for my comment on traverseLeftList/traverseRightList.

core/src/main/java/fj/data/Either.java
amarpotghan: @mperry both resides on two different classes, i.e RightProjection and LeftProjection. client has to specify which projection he has to traverse, e.g e.right().traverseList(z -> ...). I think we should have `traverseLeftList` and `traverseRightList` in Either class.

amarpotghan: @mperry that will be true for all other applicatives in Either. I think we shall avoid these translators till somebody really cares about usability. What do you think?

core/src/main/java/fj/data/List.java
```diff
@@ -563,7 +562,57 @@ public final void foreachDoEffect(final Effect1<A> f) {
     return bind(c);
   }
 
-  /**
+    /**
+     * Traverse through the List with a might-fail function.
+     *
+     * @param f The function that might fail.
+     * @return  none if applying f fails to any element of the list or f mapped list in some .
+     */
```
mperry: I would not describe the function f as a might fail function.  I would suggest `Traverses through the List with the given function`.  I checked the doc for Option.bind to check what it did.

core/src/main/java/fj/data/Either.java
amarpotghan: I have added translators for other applicatives too. Just worried if we really want to provide this convenience. I mean e.right().traverseList(z -> ...) is almost as good as e.traverseListRight(z -> ..)

core/src/main/java/fj/data/List.java
```diff
@@ -563,7 +562,57 @@ public final void foreachDoEffect(final Effect1<A> f) {
...
+    public <B> Option<List<B>> traverseOption(final F<A, Option<B>> f) {
+        F2<A, Option<List<B>>, Option<List<B>>> f2 = (a, as) -> map2Option(List::cons, f.f(a), as);
+        return foldRight(f2, Option.some(List.<B>nil()));
+    }
+
+    private static <A, B, C> Option<List<C>> map2Option(F2<A, List<B>, List<C>> f, final Option<A> opt1, final Option<List<B>> opt2) {
+        return opt1.bind(o1 -> opt2.map(o2 -> f.f(o1, o2)));
+    }
+
```
mperry: I would suggest to rename these variables, e.g. on line 372 `as` represent `Option<List<B>>`.  Why not use `obs` or `olb` here?

core/src/main/java/fj/data/List.java
```diff
@@ -563,7 +562,57 @@ public final void foreachDoEffect(final Effect1<A> f) {
...
+     * Traverse through the List with a function that might fail with reporting an error.
+     *
+     * @param f The function that produces Either value.
+     * @return  error in left if applying f fails to any element of the list or f mapped list in right.
+     */
+    public <B, E> Either<E, List<B>> traverseEither(final F<A, Either<E, B>> f) {
+        F2<A, Either<E, List<B>>, Either<E, List<B>>> f2 = (A a, Either<E, List<B>> as) -> List.<B, B, B, E>map2Either(List::cons, f.f(a), as);
+        return foldRight(f2, Either.<E, List<B>>right(List.<B>nil()));
+    }
+
```
mperry: I think I have a better solution
```
    public <E, B> Either<E, List<B>> traverseEither(F<A, Either<E, B>> f) {
        return foldRight(
                (a, acc) -> acc.right().bind(bs -> f.f(a).right().map(b -> List.cons(b, bs))),
                Either.right(List.nil())
        );
    }

```

core/src/main/java/fj/data/List.java
```diff
@@ -563,7 +562,57 @@ public final void foreachDoEffect(final Effect1<A> f) {
...
+    public <B> P1<List<B>> traverseP1(final F<A, P1<B>> f){
+        F2<A, P1<List<B>>, P1<List<B>>> f2 = (A a, P1<List<B>> ps) -> f.f(a).bind(b -> ps.map(bs -> bs.cons(b)));
+        return foldRight(f2, P.p(List.<B>nil()));
+    }
+
+    public <B> IO<List<B>> traverseIO(F<A, IO<B>> f) {
+        F2<A, IO<List<B>>, IO<List<B>>> f2 = (a, acc) ->
+                IOFunctions.bind(acc, (bs) -> IOFunctions.map(f.f(a), b -> bs.cons(b)));
+        return this.foldRight(f2, IOFunctions.unit(List.<B>nil()));
+    }
```
mperry: Now I have looked further down, this looks the same as my way above, but all on the one line to avoid having to specify the types.

core/src/main/java/fj/data/List.java
```diff
@@ -1035,7 +1084,7 @@ public final boolean exists(final F<A, Boolean> f) {
    * @return a list of all the items in this list that do not appear in the given list.
    */
   public final List<A> minus(final Equal<A> eq, final List<A> xs) {
-    return removeAll(compose(Monoid.disjunctionMonoid.sumLeft(), xs.mapM(curry(eq.eq()))));
+    return this.removeAll(compose(Monoid.disjunctionMonoid.sumLeft(), xs.mapM(Function.<A, A, Boolean>curry(eq.eq()))));
   }
```
mperry: From the diffs, I can't tell why this change is necessary.  Comments?

core/src/main/java/fj/data/List.java
```diff
@@ -1497,7 +1546,7 @@ private void tail(final List<A> tail) {
    * @return A list of the given value replicated the given number of times.
    */
   public static <A> List<A> replicate(final int n, final A a) {
-    return n <= 0 ? List.<A>nil() : replicate(n - 1, a).cons(a);
+    return n <= 0 ? List.<A>nil() : List.<A>replicate(n - 1, a).cons(a);
   }
```
mperry: Why the need for this change?

core/src/main/java/fj/data/Option.java
```diff
@@ -728,7 +727,7 @@ public A some() {
   public static <A> Option<List<A>> sequence(final List<Option<A>> a) {
     return a.isEmpty() ?
            some(List.<A>nil()) :
-           a.head().bind(aa -> sequence(a.tail()).map(cons_(aa)));
+            a.head().bind(aa -> sequence(a.tail()).map(cons_(aa)));
   }
```
mperry: Why change this?

core/src/main/java/fj/data/Validation.java
```diff
@@ -747,7 +748,7 @@ public int hashCode() {
 
 
     public <B, C, D> Validation<List<E>, D> accumulate(Validation<E, B> v2, Validation<E, C> v3, F3<T, B, C, D> f) {
-        List<E> list = fails(List.list(this, v2, v3));
+        List<E> list = fails(list(this, v2, v3));
```
mperry: Why change this?

core/src/main/java/fj/data/Option.java
amarpotghan: that's just a mistakenly added space. :)

core/src/main/java/fj/data/Validation.java
amarpotghan: @mperry I refactored out some things as and when I touched the class, I don't think it has a effect. I changed this to remove redundant static class reference and used static import there!.

core/src/main/java/fj/data/Validation.java
```diff
@@ -816,6 +817,35 @@ public int hashCode() {
         return list.foldRight(f2, Validation.success(List.nil()));
     }
 
+    public <C> List<Validation<E, C>> traverseList(F<T, List<C>> f){
+        return isSuccess() ?
+            f.f(success()).map(c -> Validation.<E, C>success(c)) :
+            list(Validation.<E, C>fail(e.left().value()));
+    }
```
mperry: Method reference for success here, e.g. `map(Validation::success)`

core/src/main/java/fj/data/List.java
amarpotghan: agreed!, you will see that changed soon!

core/src/test/java/fj/data/List_Traverse_Tests.java
```diff
@@ -0,0 +1,44 @@
...
+            if(s.startsWith("some"))
+                return some(s);
+            else
+                return Option.none();
+        };
+
+        Option<List<String>> optStr = strings.traverseOption(f);
+        Assert.assertEquals("optStr should be some", optStr.isSome(), true);
+        Assert.assertArrayEquals(optStr.some().toArray().array(), new String[]{"some1", "some2", "some3"});
+    }
```
mperry: Would be easier to compare lists

`Assert.assertArrayEquals(optStr.some(), List.list("some1", "some2", "some3"));`

core/src/main/java/fj/data/Either.java
mperry: Ok, from the diffs I did not realise this was on the left and right project.  This is fine then.

core/src/main/java/fj/data/List.java
amarpotghan: @mperry yeah I agree unless someone else uses that map2x, there is no point having a separate method. I do remember in my first commit I had a taker for map2Option, but then I moved it to list, doesn't add value. It will make sense to have inline way as you suggested.

core/src/main/java/fj/data/List.java
amarpotghan: to make Intellij happy :). There are many places Intellij fails to infer types. not sure if we want to make Intellij happy?

mperry: Hmm, I am suprised this is the case.  I am using Intellij 14.0.2.  Yourself?

core/src/main/java/fj/data/List.java
amarpotghan: @mperry foldMOption in here is nothing but traverseOption. change in replcate makes IntelliJ happy.

mperry:
I have had enough for now.  Will check back tomorrow.  Thanks for your hard work, it is much appreciated.

core/src/main/java/fj/data/List.java
amarpotghan: Ah, 13.0.1

core/src/main/java/fj/data/List.java
mperry: Tried replicate on my Windows box, Java 1.8.0-25, Intellij 14.0.2 and no complaint.  I suggest you grab the latest community edition of Intellij, I think 14.0.3 just got released.

core/src/test/java/fj/data/List_Traverse_Tests.java
amarpotghan: it only works on Array type, anyway will use `assertThat` with hamcrest matcher

mperry: Sorry, my mistake, I meant using `assertEquals` on two lists.

core/src/main/java/fj/data/List.java
amarpotghan: I need to upgrade then

amarpotghan:
@mperry done with the changes, please have a look :)

core/src/test/java/fj/data/List_Traverse_Tests.java
```diff
@@ -0,0 +1,46 @@
...
+            if(s.startsWith("some"))
+                return some(s);
+            else
+                return Option.none();
+        };
+
+        Option<List<String>> optStr = strings.traverseOption(f);
+        Assert.assertEquals("optStr should be some", optStr.isSome(), true);
+        assertThat(optStr.some(), is(List.list("some1", "some2", "some3")));
+    }
```
mperry: Do we need the hamcrest dependency when we are just saying the two lists are equal?

core/src/main/java/fj/data/List.java
```diff
@@ -563,7 +562,54 @@ public final void foreachDoEffect(final Effect1<A> f) {
...
+     */
+    public <B> Option<List<B>> traverseOption(final F<A, Option<B>> f) {
+        return foldRight(
+                (a, obs) -> f.f(a).bind(o -> obs.map(os -> os.cons(o))),
+                Option.some(List.<B>nil())
+        );
+    }
+
+    /**
+     * Traverse through the List with a function that might fail with reporting an error.
```
mperry: Eithers are unbiased and do not imply an error.

mperry:
I think we are almost done.  Just the hamcrest issue and either javadoc to address.

amarpotghan:
Hey mark,

I did not add the hamcrest dependancy. Junit has hamcrest matchers since a
long time :-)

core/src/test/java/fj/data/List_Traverse_Tests.java
amarpotghan: Junit has it since a long ago

mperry:
Still places in the javadoc that mention fail, particularly for Either.  Go to the "files changed" tab in this PR and search for "fail".  We are almost there...

PR #82, opened 2015-01-15
Add compare() and isSingle utility methods to List.

dobesv:
These are a couple of methods I find useful.  One to check if a list has only a single element, and another to compare two lists using an Ord.

### DIFF
--- a/core/src/main/java/fj/data/List.java
+++ b/core/src/main/java/fj/data/List.java
@@ -29,7 +29,6 @@ import static fj.data.Option.some;
 import static fj.function.Booleans.not;
 import static fj.Ordering.GT;
 import static fj.Ord.intOrd;
-
 import fj.Ordering;
 import fj.control.Trampoline;
 import fj.function.Effect1;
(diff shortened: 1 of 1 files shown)


core/src/main/java/fj/data/List.java
```diff
@@ -1840,4 +1839,31 @@ private void copy() {
...
+	 */
+	public Ordering compare(Ord<A> ord, List<A> other) {
+		Ordering cmp = Ord.booleanOrd.compare(isEmpty(), other.isEmpty());
+		if(isEmpty() || cmp != Ordering.EQ)
+			return cmp;
+		cmp = ord.compare(head(), other.head());
+		if(cmp != Ordering.EQ)
+			return cmp;
+		return tail().compare(ord, other.tail());
+	}
```
mperry: You may not be aware, but there are static fields on Ord for many of the common classes
Reusing this implementation means the method body is:
```
return Ord.listOrd(ord).compare(this, other);
```

Probably best just to reuse (most of) the javadoc from Ord.listOrd.

core/src/main/java/fj/data/List.java
```diff
@@ -1840,4 +1839,31 @@ private void copy() {
...
         return Show.listShow( Show.<A>anyShow() ).show( this ).foldLeft((s, c) -> s + c, "" );
     }
+
+    /**
+     * True if the list is a singleton.  Faster than checking <code>length() == 1</code> for any
+     * list with <code>length() > 1</code>.
+     */
+	public boolean isSingle() {
+		return isNotEmpty() && tail().isEmpty();
+	}
```
mperry: If would say `True if and only if the list has one element.  Runs in constant time.`

When you mentioned singleton I thought of the singleton design pattern.  

Not completely certain about the name `isSingle`.  I tried thinking of others: hasSizeOne, isSizeOne, isOne.  To me, none are significantly better.

core/src/main/java/fj/data/List.java
dobesv: Ah I hadn't thought of using Ord.listOrd . Should I just eliminate the
compare method?

mperry: I think so.

core/src/main/java/fj/data/List.java
dobesv: Yeah, the naming is tricky.  I picked `isSingle` because it matches the singleton list constructor `List.single()`.

mperry: Ok, so lets update the javadoc for isSingle and remove the compare method and I can merge the PR.

dobesv:
I'll open a new PR with just isSingle in it, to avoid cluttering up the history.

PR #83, opened 2015-01-15
Add union to TreeMap

dobesv:


core/src/main/java/fj/data/TreeMap.java
```diff
@@ -295,4 +295,20 @@ public boolean contains(final K k) {
...
+   * @return
+   */
+	public Object setAll(TreeMap<K, V> other) {
+		TreeMap<K, V> result = this;
+		for(P2<K,V> p : other) {
+			result = result.set(p._1(), p._2());
+		}
+		return result;
+	}
+
```
mperry: Lets call this union.  This is the defition of union in Haskell, https://downloads.haskell.org/~ghc/6.12.2/docs/html/libraries/containers-0.3.0.0/Data-Map.html#v%3Aunion:

```
union :: Ord k => Map k a -> Map k a -> Map k a	
O(n+m). The expression (union t1 t2) takes the left-biased union of t1 and t2. It prefers t1 when duplicate keys are encountered, i.e. (union == unionWith const). The implementation uses the efficient hedge-union algorithm. Hedge-union is more efficient on (bigset `union` smallset).

union (fromList [(5, "a"), (3, "b")]) (fromList [(5, "A"), (7, "C")]) == fromList [(3, "b"), (5, "a"), (7, "C")]
```

Please review and let me know when this is updated.  Please update the javadoc, return type (it should return TreeMap<K, V>) and method body.

There is an opportunity to implement some of the Haskell combine methods: unionWith, unionWithKey, unions and unionsWith.

There are also lots of other methods we could add from the Haskell doc above.

dobesv: Should it left biased in that case?

mperry: Yes, only add the key value pair if the key does not exist in the first map.

mperry:
@dobesv Any news of progress on this issue?

dobesv:
@mperry OK I pushed some changes, please take a look.  I had a case where I wanted to just merge in a list of pairs, and I added that as well.

core/src/main/java/fj/data/TreeMap.java
mperry: For later reference, see "Implementing Sets Efficiently in a Functional Language" by Stephen Adams as mentioned here, https://github.com/purescript/purescript-maps/issues/1.

A current link to the paper is:

http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=77314EAFB757E19867DA32911F66400C?doi=10.1.1.48.1134&rep=rep1&type=pdf

Implementation in Haskell:

https://downloads.haskell.org/~ghc/7.6.2/docs/html/libraries/containers-0.5.0.0/src/Data-Set-Base.html#hedgeUnion

mperry:
Thanks @dobesv for your contribution.

Issue #75, opened 2015-01-10
Add equals, hashCode and toString to various classes.


mperry:
Add equals, hashCode and toString to Stream, NonEmptyList, Tree, etc. as per issue #14 and PR #15 (which added them for fj.data.List).  The discussion for this topic is at https://groups.google.com/forum/?fromgroups#!topic/functionaljava/qc5fUHOg0AU.

mperry:
Do this for:
- Array, Either, List, Option, Seq, Set, Stream, Tree, TreeMap, Validation, P's, Unit.

mperry:
Addressed with PR #86.

PR #86, opened 2015-01-16
#75: Added equals, hashCode and toString to classes P's, Array, Either, ...

mperry:
...List, Option, Seq, Set, Stream, Tree, TreeMap, Validation


mperry:
Addresses issue #75.

Issue #98, opened 2015-01-27
Implement Predicate<T> 


amarpotghan:
many times I ended up in writing following combinators, I think it would add some value if library has it :

1. F&lt;A, Bool&gt; and(F&lt;A, Bool&gt; pred1, F&lt;A, Bool&gt; pred2)
2. F&lt;A, Bool&gt; or(F&lt;A, Bool&gt; pred1, F&lt;A, Bool&gt; pred2)
3. F &lt; B, Bool &gt; comap(F&lt;B, A&gt;, F&lt;A, Bool&gt;)
4. F&lt;B, Bool&gt; notFor(F&lt;B, A&gt;, F&lt;A, Bool&gt;) [a la Haskell's Data.Function.Predicate's `isn't`]
4. F&lt;A, Bool&gt; inverse(F&lt;A, Bool&gt;)

additionally we could add:
nand, nor, xor
We can also implement Predicate2<A, B>, Predicate3<A, B, C> and so on.
thoughts?

mperry:
I think we need these, the closest thing I can see is in the Booleans class (e.g. `Booleans.not`), so perhaps that is the best place to add these.

Personally, I don't think we need a Predicate class and related arities.  Given Java uses nominative typing I think conversion is a pain and would have few advantages.

I would encourage you to do this and submit a pull request.

amarpotghan:
Now as I see through `Booleans`, definitely we don't need Predicates.

techtangents:
I would also advise against a Predicate type. We had one in our fj 
equivalent, and removed it. It was just a hassle, and no benefit over 
simply writing F<A, Boolean>.

The 'and', 'or' etc functions for F<A, Boolean> are reasonable, though.

F<A, Boolean> forms 2 Moniods, one over 'and', one over 'or'. The 'or' 
variant is also an Alternative.

F<A, Boolean> or(List<F<A, Boolean>> fs) {...}
F<A, Boolean> and(List<F<A, Boolean>> fs) {...}

amarpotghan:
@techtangents every function F<A, B> should form a monoid if B is monoid. Is my understanding correct?

techtangents:
Sounds right to me.

mperry:
This is addressed in PR #102.

PR #102, opened 2015-01-30
Implement Predicate combinators #98

amarpotghan:


amarpotghan:
@mperry can you please have a look? please provide feedback.

mperry:
On my list of todos.

amarpotghan:
:) thanks

core/src/main/java/fj/function/Booleans.java
```diff
@@ -92,6 +84,56 @@ public static boolean and(final List<Boolean> l) {
...
+    /**
+     * maps given function to the predicate function
+     * @param p predicate to be mapped over
+     * @param f function
+     * @return predicate function
+     */
+    public static <A, B>  F<B, Boolean> contramap(F<B, A> f, F<A, Boolean> p){
+        return compose(p, f);
+    }
+
```
mperry: Isn't this mapping over a function, rather than a contramap?  E.g. F1W.lift(...).map(....)

core/src/main/java/fj/function/Booleans.java
```diff
@@ -92,6 +84,56 @@ public static boolean and(final List<Boolean> l) {
...
+    /**
+     * returns inverse of contramap
+     * @param p predicate to be mapped over
+     * @param f function
+     * @return predicate function
+     */
+    public static <A, B>  F<B, Boolean> isnot(F<B, A> f, F<A, Boolean> p){
+        return compose(not, contramap(f, p));
+    }
+
```
mperry: Does this imply that contramap was misnamed and should be `is`?

core/src/test/java/fj/data/BooleansTest.java
```diff
@@ -0,0 +1,81 @@
...
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testAndAll(){
+        F<String, Boolean> f1 = a -> a.startsWith("fj");
+        F<String, Boolean> f2 = a -> a.startsWith("someOtherPackage");
+        F<String, Boolean> f3 = a -> a.length() < 20;
+
+        F<String, Boolean> f4 = Booleans.andAll(Stream.<F<String, Boolean>>stream(f2, f2, f3));
+
+        Assert.assertTrue(f4.f("someOtherPackage"));
```
mperry: Does this fail because it does not satisfy f1?

mperry: I checked out your fork and realised that f1 is not used where I expected it to be in the stream of functions.  You used f2 twice.

mperry:
In summary, need to resolve `contramap`.  Thanks for this PR. In future, do work on a branch (not master) and submit a PR.

core/src/main/java/fj/function/Booleans.java
amarpotghan: technically it's a `contramap`, because it's applying function to input of `F&lt;A, Bool&gt;` (negative position)

core/src/main/java/fj/function/Booleans.java
amarpotghan: IHMO, `contramap` is more appropriate as it is in Data.Covariant.Functor http://hackage.haskell.org/package/contravariant-0.1.2/docs/src/Data-Functor-Contravariant.html. In fact, I couldn't think of better name for `isnot` so I preferred to settle down as it was named in Data.Function.Predicate

amarpotghan:
@mperry thanks, you will see changes in a branch henceforth.

tonymorris:
I think the names are fine as is. Specifically, I think `contramap` is appropriately named.

Here are some considerations:

* Rename `isnot` to simply `not`. I only make this suggestion because it has some precedence, at least in my experience.

* If you leave `isnot` named as-is, consider adding another method `is`, which is simply an alias for `contramap`. Give the users a choice between a preferred name, if it happens to be that way.

In some older Haskell libraries, with use of contravariant functors, the name `comap` was used (where `contramap` is used here). This is an inappropriate name, since the prefix `co` has a strong precedence to mean categorical duality, where no such duality has existed. Since this time, the name `contramap` has been used as the correction and I have never seen any more appropriate name.

Unlike for invariant (exponential) mapping, where I have seen several different names, for the same type of mapping; `xmap`, `expmap`, `invmap`.

core/src/main/java/fj/function/Booleans.java
mperry: I see now, it is contramapping over the predicate, whereas I was thinking it was mapping over the first function.  It looks like these two cases are equivalent here, but given we are focused on predicates contramapping is appropriate

mperry:
Data.Function.Predicate (http://hackage.haskell.org/package/predicates-0.1/docs/Data-Function-Predicate.html) uses `is` and `isn't`, so I suggest we add `is` as an alias to contramap.

mperry:
Hi @amarpotghan, why did you close this PR?

amarpotghan:
@mperry created pull-request #104 to remove commit noise :)

PR #106, opened 2015-02-11
Fix race condition in P1.memo()

### LINKED ISSUE #105: P1.memo() sometimes returns null when accessed concurrently
Paweł Krupiński reported an issue in the group as per below (https://groups.google.com/forum/#!topic/functionaljava/nSPIBAvAWVY):

Hi.
We started noticing NullPointerExceptions in our code that accessed fj.data.Stream instances concurrently.

I was able to trace it down to P1.memo() creating a P1 that sometimes returns null from ._1() call when accessed concurrently.
Here's how to reproduce it (sometimes you need to run it multiple times) on Java 8 and functionaljava 4.2:

```
final P1<String> p1 = P.p("Bla").memo();
ExecutorService executorService = Executors.newCachedThreadPool();
for (int i = 0; i < 1000; i++) {
    executorService.submit(() -> {
        if (p1._1() == null) {
            System.out.println("Null");
        }
    });
}

executorService.shutdown();
executorService.awaitTermination(10, TimeUnit.DAYS);
```

knutwalker:
The race was when two thread encountered the uninitialized state, `a == null`.
One thread gets suspended before entering the `synchronized` block, while the other runs though it. The latter thread sets `v` to something not null. Then the first thread resumes, enters the `synchronized` block and checks if `(v == null || v.get() == null)` which is now false, and thus returns the still uninitialized `a`.

Fixes #105

### DIFF
--- a/core/src/main/java/fj/P1.java
+++ b/core/src/main/java/fj/P1.java
@@ -223,9 +223,12 @@ public abstract class P1<A> implements F0<A> {
             A a = v != null ? v.get() : null;
             if (a == null)
               synchronized (latch) {
-                if (v == null || v.get() == null)
+                if (v == null || v.get() == null) {
                   a = self._1();
-                v = new SoftReference<A>(a);
+                  v = new SoftReference<A>(a);
+                } else {
+                  a = v.get();
+                }
               }
             return a;
           }
--- a/core/src/test/java/fj/P1Test.java
+++ b/core/src/test/java/fj/P1Test.java
@@ -1,0 +1,31 @@
+package fj;
+
+import org.junit.Test;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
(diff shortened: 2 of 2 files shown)


mperry:
Thanks for your fast response @knutwalker.  I read through your reasoning with the code and believe you are correct.

zot:
The code looks to me like the caching doesn't work for a _1() will that returns null, i.e. repeated calls to memo() will repeatedly evaluate _1(), contributing to the quadratic execution time that memos were created to avoid.

Maybe the condition should only be based on v == null -- something like this?  I'm using tmp to avoid a potential race condition because I'm not absolutely positive that compiled code is guaranteed to call the constructor before the assignment.  Someone who knows this for sure can remove tmp and shorten this by a line :).

    public A _1() {
      if (v != null) return v.get();
      synchronized (latch) {
        if (v == null) {
          A a = self._1();
          SoftReference<A> tmp = new SoftReference<A>(a);
          v = tmp;
          return a;
        } else return v.get();
      }
    }

mperry:
I created #108 to deal with this.

PR #110, opened 2015-02-20
Preserve ordering function when converting to a TreeMap

dobesv:
`fj.data.TreeMap` was using the default `Comparator` instance in `toMutableMap()`.  The biggest problem with this is that the default comparator requires that keys implement `Comparable`, whereas `fj.data.TreeMap` does not require keys to implement `Comparable`.  Even when keys do implement `Comparable`, the user may have used a different ordering in their `fj.data.TreeMap` which can be preserved in the resulting `java.util.TreeMap`.

### DIFF
--- a/core/src/main/java/fj/Ord.java
+++ b/core/src/main/java/fj/Ord.java
@@ -12,6 +12,7 @@ import fj.data.Validation;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.Comparator;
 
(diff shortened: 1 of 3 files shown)


mperry:
Thanks for this @dobesv.   Sorry about the delay in merging this.  Looking at this it looks like the need to reverse the keys arises from `foldMap` on `Set` doing the right branch before the left.  I suspect this should be done in the opposite order.

PR #121, opened 2015-03-31
Add bimap to Either

jbgi:
I was surprised not to found it, am I missing something?


mperry:
This looks like the Bifunctor bimap function (https://hackage.haskell.org/package/bifunctors-3.2.0.1/docs/Data-Bifunctor.html) with the signature `Either<X, Y> bimap(final F<A, X>, F<B, Y>)`.  The either function has the signature `X either(F<A, X>, F<B, X>)`.  I think this looks fine.

PR #125, opened 2015-04-11
Fix #122

### LINKED ISSUE #122: Equals.shallowEqualsO may be wrong for non-final class
and thus all Object.equals() implementation that use it in classes that allow subclassing. Pull request to fix this is on the way.
The following test fails:
```java
  @Test
  public void equals() throws Exception {
    P1<Integer> p1a = new P1<Integer>() {

      @Override
      public Integer _1() {
        return 1;
      }
    };
    P1<Integer> p1b = new P1<Integer>() {

      @Override
      public Integer _1() {
        return 1;
      }
    };

    org.junit.Assert.assertTrue(p1a + " and " + p1b + " should be equal by Object.equals", p1a.equals(p1b));
  }
```

jbgi:
This is an alternate take at resolving #122. Better performances but more casting.

### DIFF
--- a/core/src/main/java/fj/P1.java
+++ b/core/src/main/java/fj/P1.java
@@ -254,10 +254,10 @@ public abstract class P1<A> implements F0<A> {
 		return Show.p1Show(Show.<A>anyShow()).showS(this);
 	}
 
-    @SuppressWarnings("unchecked")
     @Override
+    @SuppressWarnings("unchecked")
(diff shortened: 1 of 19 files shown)


jbgi:
Rebuild this please.

mperry:
I merged PR #123 as I think it is clearer and more general.  As a result, I am going to close this PR without merging.

Issue #108, opened 2015-02-16
Caching issue when P1.memo is null


mperry:
See the comment by @zot at https://github.com/functionaljava/functionaljava/pull/106#issuecomment-74366510.

zot:
I'm not positive that the generated bytecode invokes the constructor before the assignment to v.  This code is safe in the case that it invokes the constructor after the assignment (since the first part of the code doesn't use synchronization):

    public A _1() {
      if (v != null) return v.get();
      synchronized (latch) {
        if (v != null) return v.get(); //check again here in case v was assigned asynchronously
        A a = self._1();
        SoftReference<A> tmp = new SoftReference<A>(a);
        v = tmp;
        return a;
      }
    }

gneuvill:
What do you guys think of https://gist.github.com/gneuvill/d5db376a266bf3af004c ?

I discern the following advantages of it : 
  - null safety
  - efficiency (neither locks nor synchronization)

There is one drawback though, but I can't rule on its severity : the memoized P1 may be evaluated more than once in a highly concurrent environment. But the test figuring in the gist seems to show that the number of evaluation stays negligible in comparison with the number of call to _1() (for example, 2 or 3 evaluation for 10.000.000 calls on my machine). That concern is obviously non existent in a sequential context.

P.S : the code of the memo method is copy/pasted from AtomicReference.updateAndGet available from java 8 on.

jbgi:
I am not sure I understand the rational for using SoftReference in all the implementations I have seen. Looks to me that this may lead to calling the original p1 multiple times (not what the doc says).
What about using this idiom:
https://github.com/javafp/functionaljava/commit/42cba51748c042ffaeb373d27be9185343ab90a5
?
(adapted from https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Suppliers.java)

zot:
Why the extra a variable and return statement?  Isn't OK?

@Override
        public A _1() {
          // A 2-field variant of Double Checked Locking.
          if (delegate != null) {
            synchronized (this) {
              if (delegate != null) {
                value = delegate.f();
                delegate = null;
              }
            }
          }
          return value;
        }

jbgi:
This is because access to local variable is a tiny bit faster than access to instance variable.

zot:
I think readability is more important in this case but besides that there
are some issues that might make this type of micro-optimization not so
effective or beneficial.


Is accessing a local variable really a tiny bit faster than accessing an
[...]

PR #127, opened 2015-04-14
Alternate take at fixing P1.memo cache issues #108 and #106

jbgi:
avoid repetitive locks in case of null returned by delegate.


mperry:
I have created PR #141 which takes your 3 state byte variable and uses an Option for the same purpose.  I think this is closer to the original and how we use the library, but they look essentially the same.

jbgi:
superseded by #141.

Issue #111, opened 2015-02-27
missing functions on List - (partition, safeHead)


amarpotghan:
Hey,

I would like to see : 

1. `partition :: (a -> Bool) -> [a] -> ([a], [a])` 
2. `safeHead :: [a] -> Maybe a`

on the List type, thoughts?

ms-tg:
In Scala, is this `List#partition` and `List#headOption`?

amarpotghan:
@ms-tg yup

knutwalker:
`safeHead` or `headOption` is `List#toOption`

amarpotghan:
Ah, Sure. Thanks.

tonymorris:
+1

However, `safeHead` is better written as a `Prism`

https://hackage.haskell.org/package/lens-4.7/docs/Control-Lens-Prism.html

https://hackage.haskell.org/package/lens-4.7/docs/Control-Lens-Cons.html

Note that writing prisms, in general, in Java requires code repetition, due to lack of higher-kinds. The idea here, however, is that `_head` can perform a "safe head retrieval" as well as "safe update head" operation.

amarpotghan:
thanks, it makes sense to me as a Prism! FWIW, can repetition be tolerated?

julien-truffaut:
@tonymorris `safeHead` is not a `Prism` as you cannot recreate the original `List`. However, if you compose `cons: Prism[List[A], (A, List[A])]` with `_1: Lens[(A, B), A]` you get `safeHead`. The result of the composition is an `Optional` (in monocle) or a `Traversal` (in haskell lens if I am not mistaken).

jbgi:
so, with #130, you would get:
```java
  public static <A> Optional<List<A>, A> _head() {
    return optional(
        l -> l.toOption().toEither(nil()),
        a -> l -> l.list(nil(), h -> tail -> cons(a, tail))
        );
  }
```

tonymorris:
@julien-truffaut Yes you are right.

mperry:
Created #161 to add List#partition.

PR #130, opened 2015-04-28
Optics (adapted from Scala Monocle library)

jbgi:
This is intended to address comments by @tonymorris in #68 and should allow to implement safeHead (#111) as a prism.


nremond:
:+1:

jbgi:
@functionaljava Is there something I could do so that the pull request is more easily reviewable?
maybe removing the monomorphic subclasses?

PR #132, opened 2015-05-15
Optic instances for List and P2

jbgi:
Also simplify contructors of (monomorphic) Optional and Prism.

### DIFF
--- a/core/src/main/java/fj/P2.java
+++ b/core/src/main/java/fj/P2.java
@@ -1,8 +1,10 @@
 package fj;
 
 import static fj.Function.*;
-
+import static fj.data.optic.PLens.pLens;
 import fj.data.*;
+import fj.data.optic.Lens;
+import fj.data.optic.PLens;
 
 /**
  * A product-2.
@@ -340,6 +342,34 @@ public abstract class P2<A, B> {
    */
   public static <A, B, C> F2<A, B, C> untuple(final F<P2<A, B>, C> f) {
     return (a, b) -> f.f(P.p(a, b));
+  }
+
+  /**
+   * Polyomorphic lens targeted on _1.
+   */
+  public static <A, B, C> PLens<P2<A, B>, P2<C, B>, A, C> _1pLens() {
+    return pLens(__1(), a -> p2 -> P.p(a, p2._2()));
+  }
+
+  /**
+   * Monomorphic lens targeted on _1.
+   */
+  public static <A, B, C> Lens<P2<A, B>, A> _1Lens() {
+    return new Lens<>(_1pLens());
+  }
+
+  /**
+   * Polyomorphic lens targeted on _2.
+   */
+  public static <A, B, C> PLens<P2<A, B>, P2<A, C>, B, C> _2pLens() {
+    return pLens(__2(), b -> p2 -> P.p(p2._1(), b));
+  }
+
+  /**
+   * Monomorphic lens targeted on _1.
+   */
+  public static <A, B, C> Lens<P2<A, B>, B> _2Lens() {
+    return new Lens<>(_2pLens());
   }
 
     @Override
--- a/core/src/main/java/fj/data/List.java
+++ b/core/src/main/java/fj/data/List.java
@@ -26,11 +26,21 @@ import static fj.data.Array.mkArray;
 import static fj.data.List.Buffer.*;
 import static fj.data.Option.none;
 import static fj.data.Option.some;
+import static fj.data.optic.Optional.optional;
+import static fj.data.optic.Prism.prism;
+import static fj.data.vector.V.v;
 import static fj.function.Booleans.not;
 import static fj.Ordering.GT;
 import static fj.Ord.intOrd;
 import fj.Ordering;
 import fj.control.Trampoline;
+import fj.control.parallel.Promise;
+import fj.control.parallel.Strategy;
+import fj.data.optic.Optional;
+import fj.data.optic.PTraversal;
+import fj.data.optic.Prism;
+import fj.data.optic.Traversal;
+import fj.data.vector.V2;
 import fj.function.Effect1;
 
 import java.util.AbstractCollection;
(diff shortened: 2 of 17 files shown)


core/src/main/java/fj/P2.java
```diff
@@ -342,6 +344,34 @@ public B _2() {
     return (a, b) -> f.f(P.p(a, b));
   }
 
+  /**
+   * Polyomorphic lens targeted on _1.
+   */
+  public static <A, B, C> PLens<P2<A, B>, P2<C, B>, A, C> _1pLens() {
+    return pLens(__1(), a -> p2 -> P.p(a, p2._2()));
+  }
+
```
mperry: A dumb question: Where does the naming convention for `1pLens` come from?

jbgi: no real convention here, I'm happy to change the name. I originally wanted _1Lens to be named __1 but that method already exist.
_1pLens being the polymorphic version of _1Lens.
Or should we use __1 and __2 for the lenses? not sure what would be the name for polymorphic version then...

jbgi: for reference, the Monocle library use ```first``` and ```second``` for the lenses on scala tuple:
https://github.com/julien-truffaut/Monocle/blob/master/core/src/main/scala/monocle/std/Tuple2.scala

core/src/main/java/fj/data/List.java
```diff
@@ -609,6 +619,119 @@ public final void foreachDoEffect(final Effect1<A> f) {
...
+    };
+  }
+
+  /**
+   * monomorphic traversal
+   */
+  public static <A> Traversal<List<A>, A> _traversal() {
+    return new Traversal<>(_pTraversal());
+  }
+
```
mperry: I prefer `lensTraversal` to `_traversal` for this method.  I don't like relying on using underscore to indicate the returned value is a lens.  Take `_head` vs `lensHead` for example.  

However, take note that I am a beginner user of lens.  

This would mean the `P2._1Lens` would be renamed to `lens_1`.  The polymorphic version could be `lensp_1` or `plens_1`.

Your thoughts?

jbgi: Lens is one kind of optics, Traversal is another, so lensTraversal would be confusing. so for ```_traversal``` I see only ```List.listTraversal``` or simply ``List.traversal``` as alternative naming.
Maybe @dibblego could suggest a good naming convention for optics instances in functionaljava... :)

mperry: Ok, I was using lens as the package as a whole, like the Haskell lens package (e.g. Control.Lens.Iso).

For others learning about this information, there is lots of useful info on the Github lens library page, https://github.com/ekmett/lens.

mperry:
@jbgi I don't think this is compiling on Travis using Java 1.8.0_31, see https://travis-ci.org/functionaljava/functionaljava/builds/65232022.  Can you look into this?  Other than my misgivings with naming, if we can fix this I will merge the PR.

core/src/main/java/fj/P2.java
```diff
@@ -1,8 +1,10 @@
 package fj;
 
 import static fj.Function.*;
-
+import static fj.data.optic.PLens.pLens;
```
orionll: I'm concerned about putting optic instances in the same classes. The reason is that it will cause cyclic references between packages and classes (e.g. PLens depends on List and List depends on PLens). In Monocle there is no such problem: optic instances are located in monocle.std package. What if we use the same strategy and create fj.data.optic.std package?

Also, what do you think about creating a new artifact functionaljava-lens with all fj.data.optic stuff in it?

jbgi: That is a valid concern. However there is already many cyclic packages references in fj (eg. Monoid<->Stream). Scalaz solve cyclic references by putting all in a single package. For those reasons I considered all of fj to be a single 'package'.
I can move optic instance into their own package/artifact, the downside I can see is that it will be harder to find them. pro is that it will address @mperry concern about having a clear distinction for optics 'factory methods'.

orionll: Monoid and Stream are central entities in function programming, while lenses are not (IMO). Lenses are great, but I think optic methods in "standard" classes will confuse people that are new to functional programming.
Regarding Scalaz: putting everything into a single package led to a huge jar (10 MB) which is not good at all. I would like not to repeat this mistake in functionaljava.
Anyway, this is just my personal point view, so I would like to hear other opinions.

tonymorris: Lenses (optics) are way more important to functional programming, in a practical regard.

mperry: I'd like to see them in their own package, but don't see the need for a separate artifact.

mperry: Another reason is that instances for List regarding things like Equal, Show, etc. are in those classes, rather than List.

jbgi: @mperry so... do you suggest to put optics instances in Prism, Lens, Ptraversal and the like...? I think this could be practical...
if they are isolated in their own package, those 'orphan' optics that will be harder to find which will lower their practicality.

mperry:
I don't understand what you mean by 'those orphan optics that will be harder to find...'.  

It is a suggestion for discussion, just trying to figure out the best way forward for users and potential users.

jbgi:
What I mean is that a potential user looking for the Equal<List<A>> constructor should be able to find it either in List or in Equal. I'd like to keep this practical reasoning capability for optic constructors.

orionll:
I would also hear what @runarorama thinks about it

mperry:
@jbgi At present, the classes representing the Haskell typeclasses like Equal, Show, Ord, Hash, etc. have their typeclass instances on those classes rather than List, P1, etc.

mperry:
Given my concerns are minor, I raised #154 to deal with naming conventions and package location and will merge this issue.

PR #134, opened 2015-05-19
Fixed #133 and #135

### LINKED ISSUE #133: Effect.f(Effect0) wrongly ignore its argument

jbgi:
Effect.f(Effect0) now does not ignore its argument.
Effect.comap is now static.


mperry:
Thanks for fixing those up @jbgi

PR #139, opened 2015-05-24
Add "chain" combinator for Ord

dobesv:
A combinator for Ord instances, that applies multiple Ords to the same object, in order.

### DIFF
--- a/core/src/main/java/fj/Ord.java
+++ b/core/src/main/java/fj/Ord.java
@@ -514,4 +514,21 @@ public final class Ord<A> {
   public Comparator<A> toComparator() {
 	  return new OrdComparator();
   }
+
+  /**
+   * Given two Ord instances that compare the given type of object, chain them together so
+   * that the result of the first is returned if it is not EQ, otherwise the result
+   * of the second is returned.  This can be used to reuse an Ord between subclasses
+   * or implementations of an interface.
+   *
+   * @return An Ord instance combining the two provided Ord instances
+   */
+  public static <A> Ord<A> chain(Ord<? super A> ord1, Ord<? super A> ord2) {
+	  return ord(a -> a2 -> {
+		  Ordering cmp = ord1.compare(a,  a2);
+		  if(cmp != Ordering.EQ)
+			  return cmp;
+		  return ord2.compare(a, a2);
+	  });
+  }
 }


core/src/main/java/fj/Ord.java
```diff
@@ -514,4 +514,21 @@ public int compare(A o1, A o2) {
...
+
+  /**
+   * Given two Ord instances that compare the given type of object, chain them together so
+   * that the result of the first is returned if it is not EQ, otherwise the result
+   * of the second is returned.  This can be used to reuse an Ord between subclasses
+   * or implementations of an interface.
+   *
+   * @return An Ord instance combining the two provided Ord instances
+   */
+  public static <A> Ord<A> chain(Ord<? super A> ord1, Ord<? super A> ord2) {
```
jbgi: In my understanding, functionaljava try to avoid the use of variance annotations (and subtyping as a whole), except for methods specifically created to deal with subtyping, like fj.Function.vary(F<? super A, ? extends B>).

jbgi:
What about a 'fluent' instance method similar to ```java.util.Comparator.thenComparing(Comparator<? super T>)``` instead?

jbgi:
regarding subtyping, if B extends A, comap can be used to convert a Ord<A> to an Ord<B>.

core/src/main/java/fj/Ord.java
dobesv: Without the variance annotation you can't directly reuse an Ord for a
superclass in a subclass, I think.  This is exactly why I created this
chain function, was to reuse Ord in subclasses.

dobesv:
That's a very nice idea.

core/src/main/java/fj/Ord.java
jbgi: You cannot 'directly' reuse it but you can still use ```comap``` which should do what you want at a small cost:
```java
    Ord<Date> dateOrd = Ord.comparableOrd();
    Ord<Timestamp> timestampOrd = dateOrd.comap(t -> t);
```
functionaljava does not support variance annotation because composition is usually preferred (over inheritance) when functional-programming.

tonymorris:
FWIW, `comap` is misnamed and should be called `contramap`. The current name is an accident of history.

I agree, use `contramap`, `fmap/map` and `xmap` instead of subtyping relations.

mperry:
@jbgi and @dobesv, would you like to submit a new PR for something similar to `Comparator.thenComparing`?  I am going to close this PR.

PR #140, opened 2015-05-25
Added missing functionality for FingerTree and Seq

orionll:
There was some missing base functionality in FingerTree and Seq. I've implemented the following methods:
* FingerTree: head, last, tail, init, split, split1, lookup (lookup returned null for Deep)
* Seq: head, last, tail, init, split, update, toJavaList

Also, Seq is now Iterable and can be used in foreach loops.

### DIFF
--- a/core/src/main/java/fj/data/LazyString.java
+++ b/core/src/main/java/fj/data/LazyString.java
@@ -1,13 +1,10 @@
 package fj.data;
 
-import fj.Equal;
-import fj.F;
-import fj.F2;
+import fj.*;
+
 import static fj.Function.compose;
 import static fj.Function.curry;
 import static fj.P.p;
-import fj.P1;
-import fj.P2;
 import static fj.data.Option.none;
 import static fj.data.Option.some;
 import static fj.data.Stream.join;
@@ -108,7 +105,9 @@ public final class LazyString implements CharSequence {
    * @return The String representation of this lazy string.
    */
   public String toString() {
-    return new StringBuilder(length() + 16).append(this).toString();
+    final StringBuilder builder = new StringBuilder(length() + 16);
+    s.foreachDoEffect(c -> builder.append(c.charValue()));
+    return builder.toString();
   }
 
   /**
--- a/core/src/main/java/fj/data/Stream.java
+++ b/core/src/main/java/fj/data/Stream.java
@@ -386,13 +386,7 @@ public abstract class Stream<A> implements Iterable<A> {
    * @return A new stream after performing the map, then final join.
    */
   public final <B> Stream<B> bind(final F<A, Stream<B>> f) {
-    return map(f).foldLeft((accumulator, element) -> {
-        Stream<B> result = accumulator;
-        for (B single : element) {
-            result = result.cons(single);
-        }
-        return result;
-    }, Stream.<B>nil()).reverse();
+    return foldRight(h -> (t -> f.f(h).append(t)), nil());
   }
 
   /**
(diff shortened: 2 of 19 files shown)


core/src/main/java/fj/data/LazyString.java
```diff
@@ -108,7 +105,9 @@ public CharSequence subSequence(final int start, final int end) {
    * @return The String representation of this lazy string.
    */
   public String toString() {
-    return new StringBuilder(length() + 16).append(this).toString();
+    final StringBuilder builder = new StringBuilder(length() + 16);
+    s.foreachDoEffect(c -> builder.append(c.charValue()));
+    return builder.toString();
   }
```
mperry: I don't understand why you changed this part.

orionll: Because `StringBuilder.append(CharSequence)` calls `CharSequence.charAt`, but `LazyString.chatAt` calls `Sequence.index` which is O(n). As a result, `LazyString.toString` is O(n^2). Try to convert LazyString of 10000 characters - it will not finish in reasonable time.

mperry: :+1:

core/src/main/java/fj/data/Stream.java
```diff
@@ -386,13 +386,7 @@ public final void foreachDoEffect(final Effect1<A> f) {
...
   public final <B> Stream<B> bind(final F<A, Stream<B>> f) {
-    return map(f).foldLeft((accumulator, element) -> {
-        Stream<B> result = accumulator;
-        for (B single : element) {
-            result = result.cons(single);
-        }
-        return result;
-    }, Stream.<B>nil()).reverse();
+    return foldRight(h -> (t -> f.f(h).append(t)), nil());
   }
```
mperry: I am concerned about using O(n) stack space here using foldRight given it is a common operation.

core/src/main/java/fj/data/Stream.java
```diff
@@ -1554,9 +1548,7 @@ public A head() {
    * @param o The stream of streams to join.
    * @return A new stream that is the join of the given streams.
    */
-  public static <A> Stream<A> join(final Stream<Stream<A>> o) {
-    return Monoid.<A>streamMonoid().sumRight(o);
-  }
+  public static <A> Stream<A> join(final Stream<Stream<A>> o) { return o.bind(identity()); }
```
mperry: What was the thinking behind this change?

core/src/main/java/fj/data/Stream.java
orionll: Indeed, foldRight uses O(n) space in worst case. But in this case it doesn't, because Stream.append() is lazy!
If you are not convinced, here is an example, which fails with stackoverflow with the previous implementation and doesn't fail with the new implementation:
```java
for (int j : Stream.forever(Enumerator.intEnumerator, 0).bind(i -> Stream.stream(i, i)).take(10000)) {
  System.out.println(j); // prints 0, 0, 1, 1, 2, 2, 3, 3, ...
}
```
If you are still not convinced, scalaz uses the same implementation of bind: https://github.com/scalaz/scalaz/blob/v7.2.0-M1/core/src/main/scala/scalaz/EphemeralStream.scala#L64

core/src/main/java/fj/data/Stream.java
orionll: There are two reasons:
* Performance: `sumRight()` calls strict `foldRight()` and hence requires with O(n) stack space. What is more important is that join will not work for infinite streams. On the other hand `bind() `calls lazy `foldRight`, which does not require O(n) stack space and is OK for infinite streams.
* Code conciness o.bind(identity) is shorter :)

runarorama:
This is a good set of changes. Would be good to add tests also.

PR #141, opened 2015-05-28
Attempt to fix P1.memo cache issue #108

### LINKED ISSUE #108: Caching issue when P1.memo is null

mperry:


jbgi:
:+1: maybe it is also worth overriding memo() to return ```this```?

mperry:
Good idea. :+1:

PR #144, opened 2015-06-01
Fixed bug in FingerTree.append(). Added PropertyTestRunner

orionll:
Guys, I have a suggestion. FunctionalJava has a decent property based testing framework (fj.test), but it is not used in FunctionalJava. Instead, ScalaCheck is used. I think, we should rewrite all properties to fj.test and remove ScalaCheck dependency.

Here are some advantages of this approach:
* Test classes will be a good example of fj.test usage for programmers that use FunctionalJava.
* It will test fj.test.* classes themselves.
* Programmers will trust fj.test more.
* Scala dependency will be redundant.

Don't get me wrong. I like Scala, but FunctionalJava is a library for Java and it should use Java as much as possible.

I've implemented `PropertyTestRunner` which integrates JUnit with fj.test. All classes that are marked with `RunWith(PropertyTestRunner)` will be executed during the Gradle build. `SeqProperties` is an example of a class with properties (you can run it directly from IDEA too).

I'm open for discussion.

### DIFF
--- a/core/src/main/java/fj/test/Arbitrary.java
+++ b/core/src/main/java/fj/test/Arbitrary.java
@@ -762,6 +762,16 @@ public final class Arbitrary<A> {
   }
 
   /**
+   * Returns an arbitrary implementation for sequences.
+   *
+   * @param aa An arbitrary implementation for the type over which the sequence is defined.
+   * @return An arbitrary implementation for sequences.
+   */
+  public static <A> Arbitrary<Seq<A>> arbSeq(final Arbitrary<A> aa) {
+    return arbitrary(arbArray(aa).gen.map(array -> Seq.seq((A[]) array.array())));
+  }
+
+  /**
    * Returns an arbitrary implementation for throwables.
    *
    * @param as An arbitrary used for the throwable message.
--- a/core/src/main/java/fj/test/Arbitrary.java
+++ b/core/src/main/java/fj/test/Arbitrary.java
@@ -733,6 +733,10 @@ public final class Arbitrary<A> {
     return arbitrary(listOf(aa.gen));
   }
 
+  public static <A> Arbitrary<NonEmptyList<A>> arbNonEmptyList(final Arbitrary<A> aa) {
+    return arbitrary(Gen.listOf1(aa.gen).map(list -> NonEmptyList.fromList(list).some()));
+  }
+
   /**
    * Returns an arbitrary implementation for streams.
    *
--- a/core/src/main/java/fj/test/Rand.java
+++ b/core/src/main/java/fj/test/Rand.java
@@ -124,38 +124,15 @@ public final class Rand {
   /**
    * A standard random generator that uses {@link Random}.
    */
-  public static final Rand standard = new Rand(new F<Option<Long>, F<Integer, F<Integer, Integer>>>() {
-    public F<Integer, F<Integer, Integer>> f(final Option<Long> seed) {
-      return new F<Integer, F<Integer, Integer>>() {
-        public F<Integer, Integer> f(final Integer from) {
-          return new F<Integer, Integer>() {
-            public Integer f(final Integer to) {
-              if(from == to){
-                return from;
-              }else{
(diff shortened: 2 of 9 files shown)


jbgi:
I think this a good idea overall. some considerations that may be worth taking into account:
 1. since the merge of #131, test is now a published artefact. Should the property-based testing framework be moved into that module? (as it usually not needed at runtime)
 2. Junit as now some support for property-based testing (see "Junit theories" and https://github.com/pholser/junit-quickcheck). I don't know how good it is (looks like it rely on annotations a bit too much) but maybe we should just use that??

orionll:
1. I'm not sure. `functionaljava-tests` is a module that integrates ScalaCheck with FunctionalJava, right?
2. I've just looked at `junit-quickcheck`. Looks like an imperative enterprise library. There is no spirit of functional programming in it :)

jbgi:
1. You are probably right about junit-theories. 
2. My broader question is: should the property-based testing stuff be in a separate artifact? (IMO: yes). If we get ride of scalacheck dep then functionaljava-tests could be that artifact (but this is an operational detail).

orionll:
1. Totally agree. Modularizing is always a good idea.

core/src/main/java/fj/test/Rand.java
```diff
@@ -124,38 +124,15 @@ public Random f(final Long x) {
...
-            }
-          };
-        }
-      };
-    }
+  public static final Rand standard = new Rand(seed -> from -> to -> {
+    final int min = min(from, to);
+    final int max = max(from, to);
+    final Random random = seed.map(fr).orSome(new Random());
+    return (int) ((random.nextLong() & Long.MAX_VALUE) % (1L + max - min)) + min;
```
orionll: Let me explain why I changed these lines. There were two bugs here:
* The upper bound was exclusive, but it should be inclusive.
* It didn't work for ranges wider than `Integer.MAX_VALUE` (e.g. range `[-10; Integer.MAX_VALUE]`).

core/src/main/java/fj/test/Arbitrary.java
```diff
@@ -733,6 +733,10 @@ public StringBuilder f(final String s) {
     return arbitrary(listOf(aa.gen));
   }
 
+  public static <A> Arbitrary<NonEmptyList<A>> arbNonEmptyList(final Arbitrary<A> aa) {
+    return arbitrary(Gen.listOf1(aa.gen).map(list -> NonEmptyList.fromList(list).some()));
+  }
+
```
mperry: Is this always a list of size 1, rather than some arbitrary length?

orionll: No.
* Gen.listOf - list of arbitrary length (`[0, ∞)`).
* Gen.listOf1 - non empty list of arbitrary length (`[1, ∞)`).
* Gen.sequenceN - list of specified length.

mperry: You're right, but I had to dig through the code to convince myself this is true.  The existing javadoc should be more specific.

mperry:
I think it is a good idea to redo the ScalaCheck property tests using FunctionalJava.  

Originally the property testing framework was separate called Reductio and was merged into FunctionalJava.  I think it makes sense to split it into a separate component.  I have looked at the FJ property test classes previously and I think they matched the quickcheck paper by John Hughes (http://www.eecs.northwestern.edu/~robby/courses/395-495-2009-fall/quick.pdf).  For this reason, I suggest we call the artifact `functionaljava-quickcheck` and put this into a quickcheck directory.  The Scala and FJ property tests would live in separate components.  For some context, I originally split the ScalaCheck tests into a separate module to assist IDEs.  This made it easy to exclude the ScalaCheck tests where the IDE had poor Scala support.

I still have to do some playing with the PropertyTestRunner to understand all the details.

orionll:
`functionaljava-quickcheck` - good name, I like it.

mperry:
I am going to try to split this out now.  See #147.

PR #145, opened 2015-06-03
Fixed bug in arrayShow

JackSullivan:
I noticed that the arrayShow method called in the demo gave transposed results:

`)44,22,90,98,1078,6,64,6,42Array(`

This is a fix for that.


mperry:
Thanks for your contribution Jack.

Issue #147, opened 2015-06-04
Split quickcheck property testing into separate module


mperry:
See discussion in issue #144.

PR #148, opened 2015-06-05
Split quickcheck property testing into separate module to address #147

mperry:
Address issue #147


mperry:
@orionll  @jbgi Can you look review this PR please?

orionll:
Everything looks fine.

jbgi:
I think would be more practical if the quickcheck module provide the common Arbitrary instances directly: everyone that use fj would want them.

orionll:
@jbgi What do you mean by "directly"? There are instances for all core classes (arbInteger, arbList, ...) in the Arbitrary class.

jbgi:
oups you are right.

jbgi:
as for naming, maybe 'laws' would be more explicit than 'props'?

orionll:
I'm not sure. There are functor laws, monad laws etc. But are there any laws for List, Seq? Can be the equality `list.reverse().reverse() == list` called a law?

jbgi:
@orionll yes, property is more general than law.
LGTM

PR #151, opened 2015-06-08
Replaced SoftReference with WeakReference for better performance

orionll:
Benchmark:

```java
import com.google.common.base.Stopwatch;
import fj.data.Stream;

public class Main {
    public static void main(String[] args) {
        int to = 3000000;
        Stream<Integer> range = Stream.range(1, 3000000);

        Stopwatch sw = Stopwatch.createStarted();
        Integer sum = range.foldLeft(x -> y -> x + y, 0);
        System.out.println("First run: " + sw.stop());

        sw.start();
        sum = range.foldLeft(x -> y -> x + y, 0);
        System.out.println("Second run: " + sw.stop());
    }
}
```

The result with SoftReference:
```
First run: 13,31 s
Second run: 32,68 s
```

WeakReference:
```
First run: 219,9 ms
Second run: 388,6 ms
```

SoftReference is about 70 times slower.

I also performed a profiling session with JVisualVM. With SoftReference there are 125 collections total and garbage collection took 23 seconds! With WeakReference - 29 collections and 25 milliseconds.

### DIFF
--- a/core/src/main/java/fj/data/List.java
+++ b/core/src/main/java/fj/data/List.java
@@ -624,7 +624,7 @@ public abstract class List<A> implements Iterable<A> {
    * @return A new list that has appended the given list.
    */
   public final List<A> append(final List<A> as) {
-    return fromList(this).append(as).toList();
+    return Buffer.fromList(this).prependToList(as);
   }
 
   /**
@@ -1787,7 +1787,7 @@ public abstract class List<A> implements Iterable<A> {
      * Appends (snoc) the given element to this buffer to produce a new buffer.
      *
      * @param a The element to append to this buffer.
-     * @return A new buffer with the given element appended.
+     * @return This buffer.
      */
     public Buffer<A> snoc(final A a) {
       if (exported)
@@ -1806,10 +1806,10 @@ public abstract class List<A> implements Iterable<A> {
     }
 
     /**
-     * Appends the given buffer to this buffer.
+     * Appends the given list to this buffer.
      *
-     * @param as The buffer to append to this one.
-     * @return A new buffer that has appended the given buffer.
+     * @param as The list to append to this buffer.
+     * @return This buffer.
      */
     public Buffer<A> append(final List<A> as) {
       for (List<A> xs = as; xs.isNotEmpty(); xs = xs.tail())
@@ -1817,6 +1817,28 @@ public abstract class List<A> implements Iterable<A> {
 
       return this;
     }
+
+    /**
+     * Prepends the elements of this buffer to the given list.
+     *
+     * @param as the list to which elements are prepended.
+     */
+    public List<A> prependToList(final List<A> as) {
+      if (isEmpty()) {
+        return as;
+      } else {
+        if (exported)
+          copy();
+
+        tail.tail(as);
+        return toList();
+      }
+    }
+
+    /**
+     * Returns <code>true</code> if this buffer is empty, <code>false</code> otherwise.
+     */
+    public boolean isEmpty() { return start.isEmpty(); }
 
     /**
      * Returns an immutable list projection of this buffer. Modifications to the underlying buffer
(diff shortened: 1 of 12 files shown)


jbgi:
but with WeakReference there almost not point to memoize at all (the result can be garaged really fast). Might as well remove the call memo() from the stream implementation.
Another benchmark taking advantage of stream memoization would be useful.

orionll:
We use a cache of parsed expressions in production (`com.google.common.cache.Cache<String, ASTNode>`). We tested different implementations of cache (WeakReference-based, SoftReference-based) and also no caching at all (each expression is parsed every time). The SoftReference-based cache showed terrible performance, even worse than no caching. The WeakReference-based cache showed the best performance, and we are using it in production now.

jbgi:
I think there is no general rule, it depends of how expensive it is to re-compute cached values and how frequently you need them. Maybe we need the three memo implementation to let the user chose the most appropriate one:
 - hardMemo() (basically com.google.common.base.Suppliers.MemoizingSupplier) 
 - softMemo() (current implementation) 
 - weakMemo() (your patch) 
It is totally possible that for most use case of streams weakMemo give the best results.

jbgi:
What I meant is that we should not change the current ```memo()``` behavior (which would be an alias for ```softMemo()```) and maybe instead change Stream implementation to use a new ```weakMemo()``` method on P1.

orionll:
+1 for adding three distinct methods and removing memo().

orionll:
So, I think there should be three methods:
* memo() - based on strong references
* weakMemo() - based on weak references
* softMemo() - based on soft references

See my changes.

jbgi:
@mperry can we change memo() memory behaviour without a full deprecation cycle?

techtangents:
How about we keep memo(), but add an overload for it that passes in the 
memoization strategy.

e.g.
memo(Memo.hardReference)
memo(Memo.softReference)
memo(Memo.weakReference)

then memo() would be implemented as:

memo(Memo.weakReference)

or whichever behavior you want as default.

orionll:
I don't see any benefist of it

quickcheck/src/main/java/fj/test/Gen.java
```diff
@@ -337,15 +337,7 @@ public A f(final Rand r) {
...
   public static <A> Gen<List<A>> sequence(final List<Gen<A>> gs) {
-    return gs.foldRight(new F<Gen<A>, F<Gen<List<A>>, Gen<List<A>>>>() {
-      public F<Gen<List<A>>, Gen<List<A>>> f(final Gen<A> ga) {
-        return new F<Gen<List<A>>, Gen<List<A>>>() {
-          public Gen<List<A>> f(final Gen<List<A>> gas) {
-            return ga.bind(gas, List.<A>cons());
-          }
-        };
-      }
-    }, value(List.<A>nil()));
```
orionll: This caused StackOverflow

core/src/main/java/fj/data/List.java
```diff
@@ -625,18 +624,18 @@ public final void foreachDoEffect(final Effect1<A> f) {
    * @return A new list that has appended the given list.
    */
   public final List<A> append(final List<A> as) {
-    return fromList(this).append(as).toList();
+    return Buffer.fromList(this).prependToList(as);
```
orionll: List.append() was O(n+m). I made it O(n)

mperry:
I agree with @techtangents, I think we should provide a default memo implementation and then other options to choose from a hard, soft and weak implementation.  Whether this is separate hard, soft and weak methods or a parameter to memo does not matter too much.

mperry:
@jbgi we can change the implementation of `memo` memory implementation without deprecation.  We don't provide guarantees on the implementation, so changing the implementation when the type signature remains the same I think is fine.

mperry:
I think this issue is complete enough to merge.  I created #153 to handle hard, soft, weak and default memo implementations.

PR #155, opened 2015-06-14
Added IO example from chapter 7 of Real World Haskell

mperry:


mperry:
@orionll I chose IOW for Input/Output Wrapper.  I had done this previously for functions, e.g. F1W, F2W, which I had copied from the iterable wrapper `IterableW` which I believe was there for quite some time.

Issue #154, opened 2015-06-13
Consider naming convention and location for optic instances for #132


mperry:

orionll:
What if we put optic instances into static nested classes of corresponding classes? E.g. List.Optic.*, P1.Optic.*

mperry:
I have had a think about this @orionll and have come around to your suggestion.  I think I will do this.

jbgi:
:+1:

jbgi:
@mperry then you can remove the extraneous Lens suffix of P2 lenses.

orionll:
@jbgi _1() instead of _1Lens()?

orionll:
Also, we should think where to put optic instances for standard Java classes.

jbgi:
@mperry yes.
@orionll It makes sense to have Iso instances, eg. between java.util.Optional and Option, between java.util.Supplier and P1, and maybe for java.util.List <-> fj.data.List (with the canvas that java.util.List is mutable).
but not sure what instances are you suggesting for Boolean and Integer?

orionll:
@jbgi https://github.com/julien-truffaut/Monocle/blob/master/core/src/main/scala/monocle/std/Int.scala

mperry:
There will still need to be an indiciator of whether it is using the polymorphic or monomorphic class version, e.g. `P2.Optic._1p()`.

jbgi:
@mperry true.
@orionll yes right, we should probably do a port of all relevant optics from monocle.

orionll:
@jbgi We can put them into fj.data.Java class

PR #159, opened 2015-06-23
Moved optic instances to nested class.  Added optic example

mperry:


mperry:
Addresses #154.

orionll:
Looks fine

mperry:
I still get instability in Stream.bindStackOverflow in the scalacheck tests.  I suspect it is probably related to the Stream and P1 memo implementations.  Not looking forward to looking into this.

orionll:
I think CheckStream can be deleted, because it was already rewritten to Java in StreamProperties.

PR #166, opened 2015-07-03
Fixes #162, fix Validation.sequence and rename to sequenceReduce

### LINKED ISSUE #162: Fix Validation.sequence and rename

mperry:


mperry:
Fixes #162

mperry:
@tonymorris Is there an accepted name for this?  I called it `sequenceReduce`.

mperry:
I don't know what I was thinking at the time, but I don't like renaming this as `sequenceReduce`.  I suspect reduce is the opposite of accumulate, the function does not accumulate the error values.

Issue #168, opened 2015-07-04
Scalacheck property Stream.bindStackOverflow sometimes fails


mperry:
Looks like a Scala conversion issue, perhaps to do with the mutable Iterable used.

PR #169, opened 2015-07-04
#168: Replaced iterable with Stream.range and works reliably

mperry:


mperry:
This would occasionally fail, I assume it was the mutable iterable that created the stream.  Seams to work reliably.

PR #179, opened 2015-07-23
Some fixes

jbgi:
Fixed constructors of fixed length monomorphic traversal:
 set paramter function had incorrect type

+ some minor improvements:
 - avoid unnecessary function creation
 - render inner class static


mperry:
Thanks @jbgi

PR #181, opened 2015-08-17
Fixed bug in Buffer.copy()

orionll:
There was a bug in copy():

```java
Buffer<Integer> buf = Buffer.empty();
System.out.println(buf.snoc(1).snoc(2).snoc(3).toList());
System.out.println(buf.snoc(4).toList()); // Prints List() instead of List(1,2,3,4)
```


mperry:
Nice work spotting this @orionll.  Thanks for the fix!

PR #190, opened 2015-09-04
Fix bug in Validation.accumulate corresponding to ap4

mrkm4ntr:


mperry:
Nice spot @mrkm4ntr.  Thanks for your contriubtion.

PR #208, opened 2015-12-10
Add Void: a logically uninhabited data type. Fix #193

### LINKED ISSUE #193: Add Void: a logically uninhabited data type
from https://hackage.haskell.org/package/void-0.7.1/docs/Data-Void.html

with Void you can implements technics like http://www.haskellforall.com/2015/01/total-100-exhaustive-pattern-matching.html

The implementation would be something like:
```java
public abstract class Void {
    private Void() {throw new IllegalStateException("Void cannot be instantiated"); }
    public abstract <X> X absurd();
}
```

WDYT ?

jbgi:


tonymorris:
:+1:

techtangents:
What about java.lang.Void?

PR #211, opened 2015-12-17
fix Strings#unlines for android, java6 and 7

daneko:


mperry:
`List.forEach` looks like it was introduced in Java 8.

PR #214, opened 2016-01-30
DList (Difference List)

clinuxrulz:
Difference List is designed for efficient appending to the end of lists.

**Edit**:
Benchmark: http://pastebin.com/k2n8m9UN

Average over 10 runs...
List:  1422.8ms
Seq:   187.6ms
DList: 3.9ms

The benchmark is the average of 10 runs of appending 2000 lists each containing 100 elements.

### DIFF
--- a/core/src/main/java/fj/data/DList.java
+++ b/core/src/main/java/fj/data/DList.java
@@ -1,0 +1,50 @@
+package fj.data;
+
+import fj.F;
+import fj.P;
+import fj.control.Trampoline;
+
+/**
+ * Difference List. It converts left associative appends into right associative ones to improve performance.
+ *
+ * @version %build.number%
+ */
+public class DList<A> {
+    private final F<List<A>,Trampoline<List<A>>> appendFn;
+    
+    private DList(F<List<A>,Trampoline<List<A>>> appendFn) {
+        this.appendFn = appendFn;
+    }
+    
+    public static <A> DList<A> fromList(List<A> a) {
+        return new DList<>((List<A> tail) -> Trampoline.pure(a.append(tail)));
+    }
+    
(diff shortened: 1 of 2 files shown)


core/src/main/java/fj/data/DList.java
```diff
@@ -0,0 +1,50 @@
...
+    
+    public DList<A> snoc(A a) {
+        return this.append(DList.single(a));
+    }
+    
+    public DList<A> append(DList<A> other) {
+        return new DList<>(kleisliTrampCompose(this.appendFn, other.appendFn));
+    }
+    
+    private static <A,B,C> F<A,Trampoline<C>> kleisliTrampCompose(F<B,Trampoline<C>> bc, F<A,Trampoline<B>> ab) {
```
jbgi: Could this be useful as a public method of Trampoline?

jbgi:
:+1: probably need some javadoc though...

core/src/main/java/fj/data/DList.java
clinuxrulz: Its not really a **true** Kleisli composition, because of the use of ```Trampoline.suspend(...)```. But a bit of careful rearrangement kleisli composed could be added and used from Trampoline.

mperry:
Looks good.  We should add the benchmarks to the code somewhere.  Perhaps we need a new module for this.

clinuxrulz:
I agree with keeping the benchmarks somewhere. The first time I've seen DList on Hackage, I could not make sense of how it helped the performance.

A separate module called benchmarks would be fine. I'm not experienced enough with Gradle to do it myself.

mperry:
I created #224 for the evidence of the DList.  I will merge this PR now.

PR #216, opened 2016-01-31
Process coverage for all subprojects

gliptak:
Correct processing for coveralls.io

(also see https://github.com/kt3k/coveralls-gradle-plugin/issues/38)


mperry:
Thanks for this @gliptak, this is great.  I always wanted to get this right, but didn't know why the reports were incorrect.

PR #217, opened 2016-01-31
Correction for Java8 example Option_filter.java

samekmichal:
Corrections for few small mistakes in the Java8 Option filter example.

### DIFF
--- a/demo/src/main/java/fj/demo/Option_filter.java
+++ b/demo/src/main/java/fj/demo/Option_filter.java
@@ -23,8 +23,8 @@ public final class Option_filter {
         final Option<Integer> o8 = o5.filter(f);
         final Option<Integer> o9 = o6.filter(i -> i % 2 == 0);
 
-        optionShow(intShow).println(o4); // None
-        optionShow(intShow).println(o5); // None
-        optionShow(intShow).println(o6); // Some(4)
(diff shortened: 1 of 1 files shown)


demo/src/main/java/fj/demo/Option_filter.java
```diff
@@ -19,12 +19,12 @@ public static void main(final String[] args) {
...
+        final Option<Integer> o7 = o4.filter(f);
+        final Option<Integer> o8 = o5.filter(f);
+        final Option<Integer> o9 = o6.filter(i -> i % 2 == 0);
 
-        optionShow(intShow).println(o4); // None
-        optionShow(intShow).println(o5); // None
-        optionShow(intShow).println(o6); // Some(8)
+        optionShow(intShow).println(o7); // None
+        optionShow(intShow).println(o8); // None
+        optionShow(intShow).println(o9); // Some(4)
```
mperry: Hi @samekmichal, can you change the comment here to Some(8).  It filters on the even condition without changing the value of the some.

samekmichal: Hi @mperry, of course. My bad ...

mperry:
Thanks @samekmichal

Issue #210, opened 2015-12-15
Nested class Array.ImmutableProjection should be static


mrbackend:
Array.ImmutableProjection is both non-static and takes a type parameter. This forces the user to write `Array<TypeName>.ImmutableProjection<TypeName>` (or `Array<?>.ImmutableProjection<TypeName>`). This nested class should be static, since it doesn't use any of Array's instance fields or methods.

mperry:
Agreed.  Can you submit a PR?

mrbackend:
I will :)

mrbackend:
Note: It will break backward compatibility, though.

PR #223, opened 2016-02-05
#210 Make Array.ImmutableProjection static

mrbackend:
Made nested class Array.ImmutableProjection static. See Issue #210.

Note: This is a compatibility breaker.


mperry:
Thanks @mrbackend, I will note this in the release notes.

Issue #224, opened 2016-02-05
Create a performance module for performance based decisions


mperry:
Start with #214 on DList performance.

mperry:
I am closing this issue.  See #225 for the PR.

PR #225, opened 2016-02-05
Dlist performance

mperry:
Addresses #224.


mperry:
First attempt at performance module starting with code for difference lists by @clinuxrulz in #224.  Does this look ok @clinuxrulz?

clinuxrulz:
looks good :+1:

PR #234, opened 2016-02-15
Add JUnit for DbState

gliptak:


gliptak:
@mperry Please review. Thanks

jbgi:
also to review on ```DBState```: a small improvement to avoid losing original exception by using ```addSuppressed``` for exceptions thrown by ```rollback()```/```close()```: https://github.com/functionaljava/functionaljava/pull/238/commits/a0e5f6e046ea9d3ef77e630d4b4767a1c3e423be

jbgi:
LGMT

core/src/test/java/fj/control/db/TestDbState.java
```diff
@@ -0,0 +1,42 @@
...
+                ps.setString(1, "FOO");
+                ResultSet rs = ps.executeQuery();
+                if (rs.next()) {
+                    return Option.some(rs.getInt("ID"));
+                } else {
+                    return Option.none();
+                }
+            }
+        };
+        assertThat(writer.run(setup.bind(v -> query)).some(), is(10));
```
mperry: Should we create a constant for the value 10 and use it here and when creating the insert string?

mperry: Looks like we also need a static constructor taking a `Try1<Connection, A, SQLException>`to avoid users of DB needing to instantiate an anonymous class.  This can be done in another issue.

mperry: Should we explicitly close the statement and result set as a standard practice?

mperry:
Thanks for improving the state of the database code.  This is an area that needs to get more attention.

core/src/test/java/fj/control/db/TestDbState.java
gliptak: I made the changes pointed out above.

gliptak: I'm quite sure how the Try1 is to be used ...

core/src/test/java/fj/control/db/TestDbState.java
```diff
@@ -0,0 +1,53 @@
...
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
+public class TestDbState {
+    @Test
+    public void testWriter() throws SQLException {
+        int TEN = 10;
+        DbState writer = DbState.writer(DbState.driverManager("jdbc:h2:mem:"));
+
+        DB<Unit> setup = new DB<Unit>() {
```
mperry: lambda syntax here?

core/src/test/java/fj/control/db/TestDbState.java
```diff
@@ -0,0 +1,53 @@
...
+            @Override
+            public Unit run(Connection c) throws SQLException {
```
mperry: lambda syntax here?

core/src/test/java/fj/control/db/TestDbState.java
```diff
@@ -0,0 +1,53 @@
...
+                Statement s = c.createStatement();
```
mperry: Statement is not closed.  See comment below.

core/src/test/java/fj/control/db/TestDbState.java
```diff
@@ -0,0 +1,53 @@
...
+                    } else {
+                        return Option.none();
+                    }
+                } finally {
+                    if (rs != null) {
+                        rs.close();
+                    }
+                    if (ps != null) {
+                        ps.close();
+                    }
```
mperry: I'm thinking in test code we want to use Apache Commons DBUtil.closeQuietly for closing database resources here and above.  This will still maintain our independence from dependencies for our artifacts.

core/src/test/java/fj/control/db/TestDbState.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.sql.*;
+
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
+public class TestDbState {
+    @Test
+    public void testWriter() throws SQLException {
+        int TEN = 10;
```
mperry: Make final.

core/src/test/java/fj/control/db/TestDbState.java
```diff
@@ -0,0 +1,53 @@
...
+                    }
+                    if (ps != null) {
+                        ps.close();
+                    }
+                }
+            }
+        };
+        assertThat(writer.run(setup.bind(v -> query)).some(), is(TEN));
+    }
+}
```
mperry: I think this is ready to pull.  I am tempted to pull as is, but since the changes suggestions are pretty simple, if you agree, we can get it done quickly.

core/src/test/java/fj/control/db/TestDbState.java
gliptak: good catch. there must be a better way ...

core/src/test/java/fj/control/db/TestDbState.java
gliptak: How do you lambda syntax applied here? Thanks

mperry: Use:
```
DB.db(c -> { 
  // run method contents here
});
```

core/src/test/java/fj/control/db/TestDbState.java
```diff
@@ -0,0 +1,55 @@
...
+                    s = c.createStatement();
+                    assertThat(s.executeUpdate("CREATE TABLE TEST(ID INT PRIMARY KEY, NAME VARCHAR(255))"), is(0));
+                    assertThat(s.executeUpdate("INSERT INTO TEST (ID, NAME) VALUES (" + TEN + ", 'FOO')"), is(1));
+                } finally {
+                    DbUtils.closeQuietly(s);
+                }
+                return Unit.unit();
+            }
+        };
+        DB<Option<Integer>> query = new DB<Option<Integer>>() {
```
mperry: Same here.

core/src/test/java/fj/control/db/TestDbState.java
gliptak: Thank you for the pointer. I'm running into issues with propagating the SQLException, so I would propose to commit as-is, and discuss in a followup PR.

mperry: I see, the db method takes a function (no exception), but the run method can throw an exception.  We probably want to pass in a Try1 instead.

mperry:
I created #266 to deal with using lambda syntax to create a DB state from a function that can throw a SQLException.

Issue #235, opened 2016-02-19
fj.test.Gen.pick is broken


mrbackend:
public Property pick() {
      return Property.property(Arbitrary.arbitrary(Gen.pick(2,List.range(0,10))), values -> {
        System.out.println(values);
        return Property.prop(true);
      });
    }

Prints

    List(0,1,2,3,5,6,7)
    List(0,2)
    List(0,2)
    List(1,2,3,4,5,6)
    List(0,1,3,4,5,6,7,8)
    ...

It does not pick two arbitrary elements. Am I mistaken in how this function should work?

mrbackend:
I need this to test my Trie, so I'll fix it unless I have misunderstood what's the expected behavior.

mrbackend:
This one actually needs two questions answered:

1. Should pick() generate combinations or permutations?
2. With or without replacement? (I.e. can a generated List contain several instance of the same element?)

jbgi:
fixed with https://github.com/functionaljava/functionaljava/commit/5c6a4d19c5bc74c4a5838dc643c3cc63a5560371

PR #236, opened 2016-02-27
#235 fj.test.Gen.pick is broken and #237 fj.test.Rand.reseed() is broken

mrbackend:
K, this is my first non-trivial contribution to any open source project, so bear with me if I don't follow protocol properly :) There's a change summary at the end of this text.

This pull request addresses the #235 issue, fixing a broken Gen.pick(). As I suggested in my comments on that issue, pick() really ought to come in four flavors: Combination/Permutation, with/without replacement. The same goes for Gen.someOf().

Re my comment on the #235 issue: Although the current implementation of pick() seems to suggest that it was supposed to be with replacement, the JavaDoc implies that it should be without replacement (i.e. the generated lists can't be larger than the source list), so my fixed version of pick() will generate combinations without replacement.

I maintain that pick() and someOf() ought to be deprecated, though. I don't like imbalances ;)

Some technical notes:

There is a shuffle algorithm in there - it is imperative, since no purely functional O(n) shuffle algorithms exist, as far as I know. It shouldn't be a problem, though, since the function is pure on the outside.

I have tried to make all the new functions stack safe, but no guarantees from my part. This whole FP paradigm is still new to me (about 1/2 year).

I noticed that GitHub's file compare util seemed to get really confused over my proposed changes, so here's a summary of them:

- Deprecated pick, calls pickCombinationWithoutReplacement for now
- Added pickCombinationWithoutReplacement
- Added pickCombinationWithReplacement
- Added pickPermutationWithoutReplacement
- Added pickPermutationWithReplacement
- Deprecated someOf, calls someCombinationWithoutReplacementOf for now
- Added someCombinationWithoutReplacementOf
- Added someCombinationWithReplacementOf
- Added somePermutationWithoutReplacementOf
- Added somePermutationWithReplacementOf
- And some private helper functions
- Unit tests

### DIFF
--- a/quickcheck/src/main/java/fj/test/Gen.java
+++ b/quickcheck/src/main/java/fj/test/Gen.java
@@ -1,25 +1,28 @@
 package fj.test;
 
-import static fj.Bottom.error;
-import fj.Effect;
 import fj.F;
 import fj.Function;
-import static fj.Function.flip;
-import static fj.Function.curry;
 import fj.P2;
-import static fj.P2.__1;
 import fj.Unit;
-import fj.F2;
-import static fj.data.Array.array;
+import fj.control.Trampoline;
+import fj.data.Array;
 import fj.data.List;
-import static fj.data.List.nil;
-import static fj.data.List.replicate;
 import fj.data.Option;
 import fj.function.Effect1;
 
+import static fj.Bottom.error;
+import static fj.Function.curry;
+import static fj.Function.flip;
 import static fj.Monoid.intAdditionMonoid;
 import static fj.Ord.intOrd;
-
+import static fj.P.lazy;
+import static fj.P2.__1;
+import static fj.control.Trampoline.pure;
+import static fj.control.Trampoline.suspend;
+import static fj.data.Array.array;
+import static fj.data.List.cons;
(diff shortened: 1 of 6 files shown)


mrbackend:
Sorry, forgot to create separate branches for the last one (#237). Won't happen again.

jbgi:
impressive work! thanks!
:+1:
The new pick/some methods are a bit long though, what about:
 * pickCombinationWithoutReplacement -> **combinationOf**
 * pickCombinationWithReplacement -> **selectionOf**
 * pickPermutationWithoutReplacement -> **permutationOf**
 * pickPermutationWithReplacement -> **wordOf**
 * someCombinationWithoutReplacementOf -> **someCombinationOf**
 * someCombinationWithReplacementOf -> **someSelectionOf**
 *  somePermutationWithoutReplacementOf -> **somePermutationOf**
 * somePermutationWithReplacementOf -> **someWordOf**
?

mrbackend:
It's just my Java background :)

I have no objections to your suggections. I haven't seen the term "word" for permutation with replacement, though, but my theoretic background in combinatorics is pretty minimal.

jbgi:
yeah! thanks!

PR #244, opened 2016-05-11
Add some optics for Option

puffnfresh:


jbgi:
Could be worth adding the corresponding property checks in 
https://github.com/functionaljava/functionaljava/blob/master/props-core-scalacheck/src/test/scala/fj/data/CheckOption.scala
as in:
https://github.com/functionaljava/functionaljava/blob/master/props-core-scalacheck/src/test/scala/fj/data/CheckList.scala#L209

:+1:

puffnfresh:
@jbgi I wanted that but didn't realise it would be so straightforward, thank you!

And thanks for the optics in Functional Java, they've been very useful at work.

jbgi:
thanks!

PR #245, opened 2016-05-12
Add "as" and "voided" IO functions

puffnfresh:


core/src/main/java/fj/data/IOFunctions.java
```diff
@@ -356,6 +356,14 @@ public B run() throws IOException {
         };
     }
 
+    public static final IO<Unit> when(final IO<Unit> io, final Boolean b) {
```
jbgi: I wonder if switching parameters would be more readable?

puffnfresh:
@jbgi I agree, updated.

jbgi:
thanks!

PR #246, opened 2016-05-13
Some more Semigroup instances

puffnfresh:


jbgi:
:+1: while you are at it, what about a monoid instance for IO? ;)
=> https://mail.haskell.org/pipermail/libraries/2014-November/024310.html

puffnfresh:
@jbgi added, and I also added a `Monoid<Unit>` cause that looked like it was missing too.

jbgi:
thanks!

PR #249, opened 2016-05-15
Fix backwards List#traverseIO

### LINKED ISSUE #248: List#traverseIO runs backwards
```java
List.list("Hello", "world").traverseIO(stdoutPrintln)
```

Will print:

```
world
Hello
```

puffnfresh:
Fixes #248


puffnfresh:
Now also consistent with the above and below functions, so hopefully this was the only "backwards" implementation.

gliptak:
Did the unit test for traverseIO have to change too?

puffnfresh:
@gliptak sadly there are only tests for traversing a List with an Option.

PR #251, opened 2016-05-16
More Semigroup additions

puffnfresh:
1. dual
2. min/max
3. lift
4. sumRight

### DIFF
--- a/core/src/main/java/fj/Semigroup.java
+++ b/core/src/main/java/fj/Semigroup.java
@@ -105,6 +105,14 @@ public final class Semigroup<A> {
     return semigroup(flip(sum));
   }
 
+  /**
+   * Lifts the semigroup to obtain a trivial monoid.
+   */
+  public Monoid<Option<A>> lift() {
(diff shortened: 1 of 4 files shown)


core/src/main/java/fj/Semigroup.java
```diff
@@ -97,6 +98,27 @@ public A multiply1p(int n, A a) {
   }
 
   /**
+   * Sums the given values with right-fold.
+   */
+  public A sumRight(final NonEmptyList<A> as) {
```
jbgi: what about naming it ```sumNel```? associativity make sumRight and sumLeft equivalent, and thus using foldLeft1 would be more performant...

jbgi: @puffnfresh WDYT?

puffnfresh: @jbgi I agree. I will update this.

jbgi: sorry, I did a merge of my branch and you will need to rebase now...

puffnfresh:
@jbgi rebased and fixed!

PR #252, opened 2016-05-23
Fix eager IOW.bind

fakraemer:
IOW.bind was eagerly running its IO, changed this syntax interface to just reuse IOFunctions.

### DIFF
--- a/core/src/main/java/fj/data/IOW.java
+++ b/core/src/main/java/fj/data/IOW.java
@@ -25,20 +25,11 @@ public class IOW<A> implements IO<A> {
         return io.run();
     }
 
-    public <B> IOW<B> map(F<A, B> f) {
-        return lift(() -> f.f(io.run()));
-    }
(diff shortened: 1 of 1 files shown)


jbgi:
:+1: well spotted!
You may want to associate your Atlassian email with your github account so that the commit properly link to it.

fakraemer:
Thanks, I forgot that I committed with my work config. Associated the account now. Do I need to worry about the coverall build?

jbgi:
no worry. I don't understand why coverage would decrease.

fakraemer:
@jbgi would you mind cutting a release by any chance? I'd like to make some of our code a bit more readable by using the syntax interface.

jbgi:
I also think it is time, almost: I'd like we have a fix for #257 and #251 merged before.
@mperry what do you think?

mperry:
The last release was mid Feb, so it has been 4 months.  I usually give about 1.5 weeks notice, so perhaps we should target Sat/Sun Jun 18/19.

PR #254, opened 2016-05-24
Provide an IO walkthrough example demonstrating basic composition

fakraemer:
We're trying to use functionaljava in our team, but some members are quite unfamiliar with functional programming concepts. The documentation on [www.functionaljava.org](http://www.functionaljava.org/examples-java8.html) does not contain an IO example, hence I was tasked with providing one. I stumbled over [IODemo.java](https://github.com/functionaljava/functionaljava/blob/master/demo/src/main/java/fj/demo/IODemo.java) but I think it obscures a bit what's happening, hence I strived for a simpler example with more code inlined. I realise that this might deviate a bit from the current documentation style, hence I'm happy to adapt this but I think a more trivial example might make sense given the language this library is written in and the crowd it is trying to attract (gateway drug :ghost:). I'll update the [examples-java8.html](https://github.com/functionaljava/functionaljava.github.io/blob/master/examples-java8.html) once this is merged.

### DIFF
--- a/core/src/main/java/fj/data/IOFunctions.java
+++ b/core/src/main/java/fj/data/IOFunctions.java
@@ -495,6 +495,13 @@ public class IOFunctions {
         };
     }
 
+    public static IO<Unit> stdoutPrint(final String s) {
+        return () -> {
+            System.out.print(s);
+            return Unit.unit();
+        };
+    }
+
     public static IO<LazyString> getContents() {
         Stream<IO<Integer>> s = Stream.<IO<Integer>>repeat(() -> (int) stdinBufferedReader.read());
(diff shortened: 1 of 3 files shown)


puffnfresh:
Awesome job! :+1:

jbgi:
@fakraemer is the lib open-source? I'd love to show it as a reference to my co-worker ;)
btw, did you had a look at https://github.com/derive4j/derive4j to introduce sum types/pattern matching?

fakraemer:
@jbgi you mean the project we're working on? Unfortunately it's closed-source. Will be looking to extract more example code out in the future for the functionaljava doc. I did not know about derive4j, thx! Looks neat, @puffnfresh is already looking through it to see what is feasible for us to use.

PR #258, opened 2016-05-26
Hard memo memory reduction

clinuxrulz:


core/src/main/java/fj/P1.java
```diff
@@ -268,7 +268,7 @@ public final A f() {
 	}
 
     static class Memo<A> extends P1<A> {
-      private final P1<A> self;
+      private volatile P1<A> self;
```
jbgi: ```self``` need not be ```volatile``` because the memory barrier is done via ```initialized```.
But thinking of it, you could let ```self``` be ```volatile``` and remove ```initialized``` altogether, replacing the ```!initialized``` check by ```self != null```. WDYT?

clinuxrulz: Sounds good to me.

PR #262, opened 2016-06-04
Add tests for fj.Class

gliptak:


core/src/test/java/fj/ClassTest.java
```diff
@@ -0,0 +1,53 @@
...
+        assertThat(l.length(), is(0));
+    }
+
+    @Test
+    public void testTypeParameterTree() {
+        Class c = Class.clas(Option.none().getClass());
+        Collection<Type> coll = c.classParameters().toCollection();
+        for (Type t: coll) {
+            Class.typeParameterTree(t);
+        }
```
mperry: Do we want to assert something in this method?

gliptak: updated

PR #268, opened 2016-06-07
Exclude demo subproject from Coveralls

gliptak:
I'm not sure this excludes, I couldn't get coveralls generated for my fork.


mperry:
The Travis build looks like it has excluded the demo module.

Issue #259, opened 2016-05-30
Make State stack safe


mrbackend:
I have a 90% finished (some tests need a clean-up) version of a stack safe State. Will create a PR when finished.

mperry:
Addressed in #275.

PR #270, opened 2016-06-07
Stack safe State

mrbackend:
Issue #259.

As the original file was TAB-indented, and it is my understanding that we generally use SPACE indentation, such that every single line would be marked as changed anyway, I took the liberty to reorganize the functions in a way that feels more structured, at least to me. Almost every function body is touched anyway.

And, even though the State Monad is well known among FP-ers, I think I should add some more Javadoc.

### DIFF
--- a/core/src/main/java/fj/data/State.java
+++ b/core/src/main/java/fj/data/State.java
@@ -1,120 +1,119 @@
 package fj.data;
 
-import fj.*;
+import fj.F;
+import fj.P2;
+import fj.Unit;
+import fj.control.Trampoline;
 
-import java.util.*;
-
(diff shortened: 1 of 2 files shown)


mrbackend:
"This branch has conflicts that must be resolved." What did I do wrong here?

jbgi:
looks like the file has been modified since you branch out.
I think you can do a
```git pull --rebase https://github.com/mrbackend/functionaljava.git stack-stafe-state```
then solve conflicts via eg.
```git mergetool -t kdiff3```
then push again.

jbgi:
closed in favor of #275.

PR #271, opened 2016-06-07
Remove use of Arbitrary wrapper

charleso:
This might be a little controversial. I can see how this ended up the way that it is. In particular it kinda makes sense in ScalaCheck to have `Arbitrary` distinct from `Gen`. Maybe. But without type-classes `Arbitrary` serves zero purpose.

This change is definitely an API breaking one, and depending on how seriously you take that I can understand if this isn't worth the pain. You could probably alleviate some of the source compatibility by keeping in an `arbitrary` function that just returns the `Gen` and/or sticking a redundant `gen` field on `Gen`, but that doesn't help binary compatibility which is probably the main concern.

Thoughts?


jbgi:
I have been thinking the same thing. Even in the scala world, the new https://github.com/scalaprops/scalaprops library by @xuwei-k only use ```Gen```.

Now it is a breaking change, but the API as it is now is relatively new, so I believe not many users will be impacted, provided the migration is straightforward.

charleso:
The migration could almost be expressed as a regex. :)

I do also wonder about the now redundant arb terminology,  or even the Arbitrary class, but I'm less sure whether it's worth changing all that as well. It's vaguely nice having the close mapping to QuickCheck and ScalaCheck.

puffnfresh:
:+1:

mrbackend:
So, there will be Gen and Coarbitrary, or should the latter be named something else?

charleso:
> So, there will be Gen and Coarbitrary, or should the latter be named something else?

I was going to leave a comment about keeping parity where possible with QuickCheck/ScalaCheck, but then realised the latter [already has CoGen](https://github.com/rickynils/scalacheck/blob/master/src/main/scala/org/scalacheck/Cogen.scala) as well as [scalaprops](https://github.com/scalaprops/scalaprops/blob/master/gen/src/main/scala/scalaprops/Cogen.scala). So unless anyone has any objections I'm happy to rename `Coarbitrary` while I'm here.

charleso:
I've done as a separate commit, happy to roll back if people prefer.

mperry:
A couple of comments:
- @jbgi I thought the API was old, so I am guessing lots of users are affected
- since the library closely match Hughes' Quickcheck paper, it was more easily accessible.  This change will make it harder to come from the paper, but makes usage simpler.

Regardless, I think the change makes sense.

tonymorris:
I agree, delete `Arbitrary`. It is from a time when the question, "how to do type-classes in java?" was not answered. We now know the answer (you can't). Therefore, it serves no purpose to have `Arbitrary`.

PR #272, opened 2016-06-08
Add test for IO

gliptak:


gliptak:
@mperry Please review. Thanks

core/src/test/java/fj/IOTest.java
```diff
@@ -0,0 +1,34 @@
...
+    @Test
+    public void testLift() throws IOException {
+        final IO<String> readName = () -> new BufferedReader(new StringReader("foo")).readLine();
+        final F<String, IO<Unit>> upperCaseAndPrint = F1Functions.<String, IO<Unit>, String>o(this::println).f(String::toUpperCase);
+        final IO<Unit> readAndPrintUpperCasedName = IOFunctions.bind(readName, upperCaseAndPrint);
+        readAndPrintUpperCasedName.run();
+        assertThat(result, is("FOO"));
+    }
+
+    public IO<Unit> println(final String s) {
```
mperry: Would it make sense to make this return the string, rather than assign it to the instance variable `result`.

Longer term, perhaps we need a bind method like: `IO<A> bindLeft(IO<A> io, F<A, IO<B>> f)` which binds, but returns the a from the left argument.  But this would be another issue.  I would have to check the Haskell doc to see if this exists.

gliptak: This is a way to mock the ```println``` call. In general ```print``` return ```Unit```

mperry: I know.  I thought it would be better to return `IO<String>`and return s here.  Then in the method above you can replace result with `readAndPrintUpper....` and remove the variable result.  This would improve referential transparency.

PR #274, opened 2016-06-12
Implemented Equal, Hash, Arbitrary, and Shrink for Natural as well as…

bakeemawaytoys:
… the equals, hashCode and toString methods on Natural.  Also fixed the subtract method on Natural as well as the Javadoc on the Natural Semigroup for addition.


core/src/main/java/fj/data/Natural.java
```diff
@@ -286,4 +291,20 @@ public static Natural sum(final List<Natural> ns) {
...
+
+
+  @Override
+  public int hashCode() {
+    return Hash.naturalHash.hash(this);
+  }
+
+  @Override
+  public boolean equals(final Object that) {
+    return Equal.equals0(Natural.class, this, that, Equal.equal(l -> r -> l.value.equals(r.value)));
```
jbgi: last argument should be ```Equal.naturalEqual```

quickcheck/src/main/java/fj/test/Arbitrary.java
```diff
@@ -1337,4 +1338,10 @@
     return aa.bind(ab, ac, ad, ae, af, ag, ah,
         a -> b -> c -> d -> e -> f -> g -> h -> p(a, b, c, d, e, f, g, h));
   }
+
+  /**
+   * An arbitrary implementation for naturals.
+   */
+  public static Gen<Natural> arbNatural = arbLong.map(Longs.abs).map(Natural::natural).map(o -> o.some());
```
jbgi: this will fail for ```Long.MIN_VALUE``` since ```Math.abs``` is the identity in this case. better filter it out.

jbgi: I think it would be nice if the generator could generate some values above ```Long#MAX_VALUE```. But I'm not sure what is the best strategy for doing that... WDYT?

bakeemawaytoys: I agree.  In fact, my first implementation of Gen and Shrink for Natural used BigInteger instead of Long.  The problem that I ran into is that the BigInteger Shrink implementation is broken.  There is a call to System.arraycopy in the implementation that throws an exception because the source array is byte[] and the destination array is Byte[].  Once I fixed that, I did some testing of the BigInteger shrink and found that it doesn't appear to shrink correctly.

For example
```
Shrink.shrinkLong.shrink(1000L).take(100).toList();
//Produces List(0,0,500,750,875,938,969,985,993,997,999)
```
versus
```
Shrink.shrinkBigInteger.shrink(BigInteger.valueOf(1000L)).take(100).toList();
//Produces List(0,0,0,512,0,0,0,512,2,2,2,514)
```

I'm not entirely sure how the current implementation is supposed to work so I made an  implementation of Shrink.shrinkBigInteger that uses the exact same logic as Shrink.shrinkLong.  It appears to work correctly although I'm not sure if that is a good approach.  At this point, I think the two options are to go with the Natural Shrink and Gen implementations that uses Long or fix the BigInteger Shrink implementation and then switch the Natural implementations to use BigInteger.

mperry: @bakeemawaytoys In #276 I put some comments on what other libs do.  Looks to me we should use the same logic.

mperry:
I propose we merge this and then fix the shrinkBigInteger (#276) and improve the gen and shrink for natural in another issue (that is not yet created).

jbgi:
merged with 09b6713407a2f3e2d1eb2c2ebd72c402a8797262 with some minor cleanup (use of contramap, added missing final modifier)

jbgi:
@bakeemawaytoys you might want to associate the email address used for the commit (johnson.1922@osu.edu) with your github account, so that the commit is linked to your account.

PR #279, opened 2016-06-18
Add a priority queue based on finger trees

mperry:


mperry:
@jbgi Can you review this please?

core/src/main/java/fj/data/List.java
```diff
@@ -962,7 +962,7 @@ else if (tail().isEmpty())
             final A x = xs.head();
             final A y = ys.head();
 
-            if (o.isLessThan(x, y)) {
+            if (o.isLessThan(x, y) || o.eq(x, y)) {
```
jbgi: Adding a ```isLessThanOrEqualTo``` to ```Ord``` would be useful to avoid doing the comparison twice.

core/src/main/java/fj/Monoid.java
```diff
@@ -402,6 +402,16 @@ public A join(final Iterable<A> as, final A a) {
   }
 
   /**
+   * A monoid for the maximum of two integers.
+   */
+  public static final Monoid<Integer> intMaxMonoid = monoid(Semigroup.intMaximumSemigroup, Integer.MIN_VALUE);
```
jbgi: Do you have a use case for this? It is a valid instance but does not feel very mathematical as it rely on technicalities of integer representation limitations.
If I'm  looking for a max value, I rarely want to default to ```Integer.MIN_VALUE```: I prefer to be explicit about the default and use the semigroup.

core/src/main/java/fj/data/List.java
mperry: Agreed.  I have pushed this change.

core/src/main/java/fj/Monoid.java
mperry: I'm using this monoid to annotate the nodes of the finger tree (the measured values) with integers and the max operation means the finger tree represents a priority queue.

The existing finger tree implementation uses a monoid to annotate the node, so I using it for that.

props-core/src/test/java/fj/data/fingertrees/FingerTreeTest.java
```diff
@@ -36,33 +34,71 @@ void validateSize(List<Integer> list) {
...
     }
 
+    @Test
+    public void stream() {
+        FingerTree<Integer, Integer> ft = midSeq();
+        out.println(ft);
+        Stream<Integer> s = ft.toStream();
+        System.out.println(s.toList());
+
+        out.println(midPriorityQueue());
```
jbgi: missing asserts?
(same for ```split```)
Also commented code and ```println``` calls should probably be removed.

props-core/src/test/java/fj/data/properties/PriorityQueueProperties.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Sorting a list returns the same as putting the list into a priority queue and getting the queue as a list.
+     */
+    public Property sorted() {
+        return property(arbPriorityQueueIntegerString, pq -> {
+            List<P2<Integer, String>> expected = pq.toList().sort(Ord.p2Ord1(Ord.intOrd.reverse()));
+            List<P2<Integer, String>> actual = pq.toList();
+            assertThat(actual, equalTo(expected));
```
jbgi: assert should be removed since it will be tested by the property.

core/src/main/java/fj/data/PriorityQueue.java
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * If the tree is not empty, returns the node with highest priority otherwise returns nothing.
+     */
+    public Option<P2<K, A>> top() {
+        K top = ftree.measure();
+        P2<FingerTree<K, P2<K, A>>, FingerTree<K, P2<K, A>>> p = ftree.split(k -> equal.eq(top, k));
+        return p._2().headOption();
+    }
+
+    public List<P2<K, A>> topN() {
```
jbgi: a javadoc would be nice, something like "Return all the elements at the top of the queue that have the same highest priority".

core/src/main/java/fj/data/PriorityQueue.java
```diff
@@ -0,0 +1,206 @@
...
+import fj.data.fingertrees.FingerTree;
+
+import static fj.data.List.list;
+
+/**
+ * A priority queue implementation backed by a {@link fj.data.fingertrees.FingerTree}.  The finger tree nodes are annotated with type K, are combined using a monoid of K and both the key and value are stored in the leaf.  Priorities of the same value are returned FIFO (first in, first out).
+ *
+ * Created by MarkPerry on 31 May 16.
+ */
+public class PriorityQueue<K, A> {
```
jbgi: could it be made final? (since there is only one implementation)

core/src/main/java/fj/data/PriorityQueue.java
```diff
@@ -0,0 +1,206 @@
...
+import fj.P;
+import fj.P2;
+import fj.P3;
+import fj.Show;
+import fj.data.fingertrees.FingerTree;
+
+import static fj.data.List.list;
+
+/**
+ * A priority queue implementation backed by a {@link fj.data.fingertrees.FingerTree}.  The finger tree nodes are annotated with type K, are combined using a monoid of K and both the key and value are stored in the leaf.  Priorities of the same value are returned FIFO (first in, first out).
```
jbgi: The line is a bit long in the editor ;)

core/src/main/java/fj/data/PriorityQueue.java
```diff
@@ -0,0 +1,206 @@
...
+    private PriorityQueue(Equal<K> e, FingerTree<K, P2<K, A>> ft) {
+        equal = e;
+        ftree = ft;
+    }
+
+    /**
+     * Creates a priority queue from a finger tree.
+     */
+    public static <K, A> PriorityQueue<K, A> priorityQueue(Equal<K> e, FingerTree<K, P2<K, A>> ft) {
+        return new PriorityQueue<K, A>(e, ft);
```
jbgi: nitpicking: could be ```new PriorityQueue<>(```

core/src/main/java/fj/data/PriorityQueue.java
```diff
@@ -0,0 +1,206 @@
...
+     */
+    public static <K, A> PriorityQueue<K, A> empty(Monoid<K> m, Equal<K> e) {
+        return priorityQueue(e, FingerTree.empty(m, (P2<K, A> p) -> p._1()));
+    }
+
+    /**
+     * An empty priority queue with integer priorities.
+     */
+    public static <A> PriorityQueue<Integer, A> emptyInt() {
+        return priorityQueue(Equal.intEqual, FingerTree.empty(Monoid.intMaxMonoid, (P2<Integer, A> p) -> p._1()));
```
jbgi: could be ```return empty(Monoid.intMaxMonoid, Equal.intEqual);```

core/src/main/java/fj/data/PriorityQueue.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Creates an empty priority queue.
+     *
+     * @param m A monoid to combine node annotations.
+     * @param e A value to compare key equality.
+     */
+    public static <K, A> PriorityQueue<K, A> empty(Monoid<K> m, Equal<K> e) {
+        return priorityQueue(e, FingerTree.empty(m, (P2<K, A> p) -> p._1()));
```
jbgi: could be ```priorityQueue(e, FingerTree.empty(m, P2.__1()))```

core/src/main/java/fj/data/PriorityQueue.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Maps the values in each node with function f.
+     */
+    public <B> PriorityQueue<K, B> map(F<A, B> f) {
+        return priorityQueue(equal,
+                ftree.map(p2 -> p2.map2(a -> f.f(a)),
+                FingerTree.measured(ftree.measured().monoid(), (P2<K, B> p2) -> p2._1()))
+        );
```
jbgi: could be
```java
        return priorityQueue(equal,
                ftree.map(P2.map2_(f),
                FingerTree.measured(ftree.measured().monoid(), P2.__1()))
        );
```

core/src/main/java/fj/data/PriorityQueue.java
```diff
@@ -0,0 +1,206 @@
...
+        K top = ftree.measure();
+        P2<FingerTree<K, P2<K, A>>, FingerTree<K, P2<K, A>>> p = ftree.split(k -> equal.eq(k, top));
+        FingerTree<K, P2<K, A>> right = p._2();
+        return right.isEmpty() ? this : priorityQueue(equal, p._1().append(right.tail()));
+    }
+
+    /**
+     * Returns a tuple of the node with the highest priority and the rest of the priority queue.
+     */
+    public P2<Option<P2<K, A>>, PriorityQueue<K, A>> dequeueTop() {
```
jbgi: should it be named "topDequeue" to match result order?

jbgi:
@mperry I opened a PR on your branch with some of the suggested changes: https://github.com/mperry/functionaljava/pull/2

props-core/src/test/java/fj/data/fingertrees/FingerTreeTest.java
mperry: I was playing with the code to understand what was happening.  Removed.

props-core/src/test/java/fj/data/properties/PriorityQueueProperties.java
mperry: Yep.

core/src/main/java/fj/data/PriorityQueue.java
mperry: There is a bunch of these where I later added convenience methods and did not go back and improve the original code.  Thanks for picking this up.

core/src/main/java/fj/data/PriorityQueue.java
mperry: I have a bad tendency to reimplement, rather than reuse, these small functions.

mperry:
I have merged these and addressed each of these issues.

Do you think it is ready to merge into the Functional Java master branch?

I am too close to the code now and find it hard to spot the errors.

jbgi:
@mperry yep, look good to me! I'm tidying up the history a bit and will push soon.

PR #280, opened 2016-06-19
Add a Hash Array Mapped Trie

mperry:


core/src/main/java/fj/data/hamt/BitSet.java
```diff
@@ -0,0 +1,195 @@
...
+    public Stream<Boolean> toStream() {
+        return Stream.fromString(Long.toBinaryString(value)).map(c -> toBoolean(c)).dropWhile(b -> !b);
+    }
+
+    public String toString() {
+        return Show.bitSetShow.showS(this);
+    }
+
+    public boolean equals(Object obj) {
+        return Equal.equals0(BitSet.class, this, obj, () -> Equal.bitSetSequal);
```
jbgi: you can use the overload of ```equals0``` that take directly the ```Equal``` instance as last parameter.

mperry:
Some things that I am concerned about:
* Should this bit set be limited to 64 bits
* IllegalArgumentException in BitSet, should this be handled with Option/Validation?
* Need to spend time lining this up to the Haskell/Scala BitSet.
* BitSet is probably slow compared to other implementations.
* I always find shifting left until the first bit is a one, making the number negative strange behaviour and am usually suprised until I have to deal with it a few times.
* Temporary methods in SeqUtil.
* Use of Seq instead of list or array for the HAMT nodes.

mperry:
@jbgi I think this stuff is good enough to use.  Even with the bit set problems, I don't want the good to be the enemy of the great.  Did you take a look earlier?  I have fixed a number of bugs, added javadoc and added properties.  There are still issues with BitSet, but I don't have the inclination to make an awesome BitSet, just one that is good enough.  The issues remaining are:

* Should BitSet this bit set be limited to 64 bits
* IllegalArgumentException in BitSet, should this be handled with Option/Validation?
* Need to spend time lining this up to the Haskell/Scala BitSet.
* BitSet is probably slow compared to other implementations.
* I always find shifting left until the first bit is a one, making the number negative strange behaviour and am usually suprised until I have to deal with it a few times.

jbgi:
I wanted to have a deeper look at the implementation and the theory but I will still be overbooked with work for the next 2 weeks.
Maybe @mrbackend or @clinuxrulz would like to do a review?

mperry:
Any interest from @mrbackend or @clinuxrulz?  I know @mrbackend did some work towards this, so he might be particularly interested.

jbgi:
next stop after HAMT: HHAMT!
=> http://arxiv.org/pdf/1608.01036.pdf
featured usage: persistent multimaps.

jbgi:
@mrbackend Would you have a go to review this? (since we are going to cut 4.7 soon)
otherwise I will do a shallow review (I have very limited previous knowledge of HAMT) this week-end before merging.

jbgi:
Superseded by #280: bitset perf improvement and some minor cleanup .

Issue #265, opened 2016-06-06
Support both java 8 and java 6 binary artifacts


mperry:
At the moment everything is backported to Java 6, but projects on Java 8 should be able to take advantage of runtime lambda support in Java 8.

jbgi:
Nowadays, I think backporting to Java7 is enough. Having to install jdk6 is a burden...

gliptak:
With https://github.com/functionaljava/functionaljava/pull/302 support for 1.8 was added.
Java6 is still targeted?
Thanks

jbgi:
With #302 only java 7 an 8 will be target. If this is a problem we can try to set 
[`retroLambdaTarget`](https://github.com/functionaljava/functionaljava/blob/master/build.gradle#L81) to `JavaVersion.VERSION_1_6` and see if it is enough to target java 6.

mperry:
I tried this and the build produces JVM 6 code.  That should be enough.  If we start targeting JVM 8+ with FJ 5 might be best just to continue to support 6 for 4.x versions, which means we need to change build.gradle to use 

```
retroLambdaTarget = JavaVersion.VERSION_1_6
```

jbgi:
done in 1faf5b9184d82cca5929747f5cfb4e67100752ab

PR #284, opened 2016-06-21
Optimizable typeclasses

jbgi:
A new design that allow to better optimize "typeclass" by allowing to override some primitive operations and also allowing alternative definitions.

The idea is that a `Ord.Definition` extends `Equal.Definition`.
same idea would apply for Semigroup/monoid.

This new design is backward compatible but will only be useful in user code if #265 is "fixed" (ie. only for jdk8 projects).

@functionaljava WDYT?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -13,7 +13,7 @@ The library is intended for use in production applications and is thoroughly tes
 Functional Java provides abstractions for the following types:
 
 * Basic Data Structures - total and partial functions, products, unit, option, unbiased and right biased unions (either and validation), void.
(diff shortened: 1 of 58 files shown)


clinuxrulz:
This design allows someone to use captureless lambas to avoid a unnecessary memory allocation during comparisions in ```Eq``` and ```Ord``` ? Seems like a great idea.

jbgi:
@clinuxrulz yes, for `Ord` and `Equal` that is the point.

jbgi:
Should be mergeable now.

PR #286, opened 2016-06-22
First class cata for either and option

jbgi:


mperry:
The methods that produce these style of first class function are often named with a postfix '_'.  E.g. Option.option_, P1.map_, etc.  I am unsure whether these should be named 'option_' and 'either_' or not.

jbgi:
@mperry good point. I was also not sure about naming the methods. I will rename them to use the suffix scheme.

PR #289, opened 2016-07-18
deprecate List#groupBy that take an implicit ordering...

jbgi:
based on Object#hashCode.
Also replaced implementation to use the slightly better Ord#hashEqualsOrd.


jbgi:
in fact the problem is a bit larger... opening another pull request.

PR #290, opened 2016-07-18
Deprecate unsafe Ord instances and List#groupBy that use them.

jbgi:


jbgi:
@mperry should those dangerous methods/instances be removed altogether (without deprecation)?

PR #298, opened 2016-09-21
Introduce Sum-types

nikita-volkov:
This PR introduces Sum Types of different arities. Sum Type is a dual of Product Type, so since the library already has the Product types, it is natural to expect the Sum types as well.

The implementation is only a sketch so far, but I feel it should be a good basis for the full-blown integration.

For more info about Sum Types, see https://en.wikipedia.org/wiki/Tagged_union.

### DIFF
--- a/core/src/main/java/fj/Function.java
+++ b/core/src/main/java/fj/Function.java
@@ -863,4 +863,38 @@ public final class Function {
       final F<A, F<B, F<C, F<D, F<E, F<F$, F<G, F<H, I>>>>>>>> f, final H h) {
     return a -> b -> c -> d -> e -> f$ -> g -> uncurryF8(f).f(a, b, c, d, e, f$, g, h);
   }
+
+  /**
+   * Unite functions into a function on a sum-type.
+   */
+  public static <result, _1, _2> F<S2<_1, _2>, result> sum(
+    F<_1, result> projection1,
+    F<_2, result> projection2
+  ) {
+    return s -> s.match(projection1, projection2);
+  }
+
+  /**
+   * Unite functions into a function on a sum-type.
+   */
+  public static <result, _1, _2, _3> F<S3<_1, _2, _3>, result> sum(
+    F<_1, result> projection1,
+    F<_2, result> projection2,
+    F<_3, result> projection3
+  ) {
+    return s -> s.match(projection1, projection2, projection3);
+  }
+
+  /**
+   * Unite functions into a function on a sum-type.
+   */
+  public static <result, _1, _2, _3, _4> F<S4<_1, _2, _3, _4>, result> sum(
+    F<_1, result> projection1,
+    F<_2, result> projection2,
+    F<_3, result> projection3,
+    F<_4, result> projection4
+  ) {
+    return s -> s.match(projection1, projection2, projection3, projection4);
+  }
+
 }
--- a/core/src/main/java/fj/S2.java
+++ b/core/src/main/java/fj/S2.java
@@ -1,0 +1,45 @@
+package fj;
+
+import fj.data.Option;
+
+/**
+ * A sum-type with 2 alternatives.
+ */
+public final class S2<_1, _2> {
+
+  private final int tag;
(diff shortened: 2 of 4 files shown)


tonymorris:
Is this not `Either` and then `Either3` and so on?

nikita-volkov:
Well, kinda. It would be more correct to say that `Either` is a sum type with 2 alternatives though. 

A name `Either3` both wouldn't make much sense linguistically or have a theory reference. OTOH, Sum is a logical dual of Product both linguistically and is named so in theory.

tonymorris:
You've clearly not used Scala then, with its `Product0`, `Product1`, `Product2`, `Product3`, … `Product22`

:)

nikita-volkov:
You didn't get me right. I don't have anything against `Product`. What I'm saying is that `Sum3` is a better name than `Either3`.

tonymorris:
I am ambivalent on the matter of naming. I thought `Either3` existed. I thought I wrote it, 10 years ago or so.

nikita-volkov:
Nope. There is none.

jbgi:
@nikita-volkov I think it is a useful addition, but it would be even more useful if we also introduce a corresponding object algebras hierarchy. Ie, something like:
```java
public interface S2<A, B> {

  interface Cases<A, B, X> {
    X _1(A a);
    X _2(B b);
  }

  <X> X s2(Cases<A, B, X> cases);

  default <C> S3<A, B, C> asS3() {
    return this::s2;
  }

  static <A, B, X> Cases<A, B, X> cases(F<A, X> _1, F<B, X> _2) {
    return new Cases<A, B, X>() {
      @Override public X _1(A a) {
        return _1.f(a);
      }

      @Override public X _2(B b) {
        return _2.f(b);
      }
    };
  }

  default Option<A> _1() {
    return s2(cases(
        Option::some,
        __ -> Option.none()
    ));
  }

  static <A,B> S2<A, B> _1(A a) {
    return new S2<A, B>() {
      @Override public <X> X s2(Cases<A, B, X> cases) {
        return cases._1(a);
      }
    };
  }

  static <A,B> S2<A, B> lazy(F0<S2<A, B>> s2) {
    P1<S2<A, B>> eval = P.lazy(s2);
    return new S2<A, B>() {
      @Override public <X> X s2(Cases<A, B, X> cases) {
        return eval.f().s2(cases);
      }
    };
  }
}
```
```java
public interface S3<A, B, C> {

  interface Cases<A, B, C, X> extends S2.Cases<A, B, X> {
    @Override
    X _1(A a);
    @Override
    X _2(B b);

    X _3(C c);
  }

  <X> X s3(Cases<A, B, C, X> cases);

// ... 
}
```
(btw, most of this could be generated via: [derive4j](https://github.com/derive4j/derive4j))

WDYT?

mperry:
I favour a general set of sum type classes using S$X, similar to the P$X set of product classes.  Something we need to address, how do we manage the existing Either class which would duplicate the S2 class?

If we can generate these classes, then let's do that as much as possible.  There is no extra value in doing this by hand, hand coding acts as an impediment to progress.

Note that the use of default methods on interfaces requires Java 8.   Is RetroLambda now mature enough to support this or do we wait until FJ 5 with Java 8 support only for this?

I think we need to move to Java 8 support only with FJ 5 ASAP after a FJ 4.7 release.

jbgi:
with retrolambda, default methods are fine as long as the interface is not designed to be extended outside the compilation unit.

gneuvill:
> I think we need to move to Java 8 support only with FJ 5 ASAP after a FJ 4.7 release

I second that.

gliptak:
With 4.7 released, can this be moved forward? Thanks

gliptak:
@nikita-volkov Could you update this PR with tests?

nikita-volkov:
@gliptak No unfortunately. I'm outta Java-ing for now.

jbgi:
@nikita-volkov I think you meant _fortunately_ (assuming you are now using Haskell)

Closing for now.

PR #300, opened 2016-09-26
some minor fixes/improvements of java8 module 

jbgi:
- use more method references.
- make Stream conversions respect laziness.
- optimize collectors.

### DIFF
--- a/java8/src/main/java/fj/data/Java8.java
+++ b/java8/src/main/java/fj/data/Java8.java
@@ -112,8 +112,11 @@ public final class Java8 {
         return Java8::Optional_Option;
     }
 
+    /**
+     * Convert an Option to {@link Optional}. Will throw a {@link NullPointerException} if the Option is some(null).
+     */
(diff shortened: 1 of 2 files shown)


java8/src/main/java/fj/data/Java8.java
```diff
@@ -28,95 +29,95 @@ private Java8() {
...
     }
 
     public static <A> F<Optional<A>, Option<A>> Optional_Option() {
-        return o -> o.isPresent() ? Option.some(o.get()) : Option.none();
+        return Java8::Optional_Option;
     }
 
     public static <A> Optional<A> Option_Optional(final Option<A> o) {
-        return Java8.<A>Option_Optional().f(o);
+        return o.option(Optional.empty(), Optional::ofNullable);
```
jbgi: Should we really map `Some(null)` to `Optional.absent()`?
It is weird as the reverse mapping of `Optional.absent()` is `None()`.
@functionaljava can I change that and throw an `IllegalArgumentException` instead?
This would allow a lawful `Iso`.

PR #305, opened 2016-12-06
Faster TreeMap.get

siro-mateos:
Hello!

TreeMap.get was using Set.split, so it was very slow on mid-sized trees because of the overhead of creating two subtrees every time, just to be discarded.

So I've added a new operation to Set, named 'peek' (i'm not native English speaker, so please feel free to use other name if you wish), wich, given some element, returns an element in the set wich equals to the one given, or None; it's just a fast loop search in the tree. Now, TreeMap can use 'peek' instead of 'split' for finding some value given a key, and it goes *a lot* faster now.

Greets!

### DIFF
--- a/core/src/main/java/fj/data/Set.java
+++ b/core/src/main/java/fj/data/Set.java
@@ -516,6 +516,129 @@ public abstract class Set<A> implements Iterable<A> {
   }
 
   /**
+   * Find element equal to the given one.
+   *
+   * @param a An element to compare with.
+   * @return Some element in this set equal to the given one, or None.
+   */
+  public final Option<A> lookup(final A a) {
+    Set<A> s = this;
+    while (true)
+      if (s.isEmpty())
+        return none();
+      else {
+        final A h = s.head();
+        final Ordering i = ord.compare(a, h);
+        if (i == LT)
+          s = s.l();
+        else if (i == GT)
+          s = s.r();
+        else
+          return some(h);
+      }
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/fj/data/Set.java
```diff
@@ -516,6 +516,30 @@ public final int size() {
   }
 
   /**
+   * Searches the set for an element equal (by the set's Ord) to the one provided. Similar to {@link #split(A)},
```
jbgi: I think you can drop the "Similar to..." off the doc.

core/src/main/java/fj/data/Set.java
```diff
@@ -516,6 +516,30 @@ public final int size() {
   }
 
   /**
+   * Searches the set for an element equal (by the set's Ord) to the one provided. Similar to {@link #split(A)},
+   * without computing any subsets.
+   *
+   * @param a An element to compare with.
+   * @return Some element in this set equal to the one provided, or None.
+   */
+  public final Option<A> peek(final A a) {
```
jbgi: I think this function should be named either:
- `lookup`: to be consistent with `TreeMap#splitLookup` (and [haskell](https://hackage.haskell.org/package/containers-0.5.8.1/docs/Data-Set.html#v:lookupLT) ).
- or `get`: to be consistent with `TreeMap#get` (but which name is kind of inconsistent with `splitLookup`/haskell)

I would prefer `lookup`.

PR #307, opened 2016-12-07
Bump coveralls-gradle-plugin to 2.7.1

gliptak:


gliptak:
Still fails to generate coveralls data ...

PR #317, opened 2017-05-10
Fix broken NonEmptyList semigroup implementation.

bakeemawaytoys:
I found a bug in the NonEmptyList semigroup implementation when I attempt to update one of my projects to use 4.7.  I've fixed the bug and added some Scalacheck test coverage for the Semigroup class and its implementations.  I also created some new implicit conversions of the Scalacheck Arbitrary to facilitate the new tests.


jbgi:
Awesome thanks!

tonymorris:
Blimey how long has that been around?

jbgi:
@tonymorris I think this is a regression introduced by a refactoring in last release. 4.6 and earlier should be fine.
We should do a release with the fix asap (I'll try this week).

PR #322, opened 2017-09-08
Reference from static field of LazyString class to Equal class removed

l1cache:
Cause it could cause deadlock on parallel static initialization.

### DIFF
--- a/core/src/main/java/fj/Equal.java
+++ b/core/src/main/java/fj/Equal.java
@@ -674,7 +674,7 @@ public final class Equal<A> {
   /**
    * An equal instance for lazy strings.
    */
-  public static final Equal<LazyString> eq = streamEqual(charEqual).contramap(LazyString.toStream);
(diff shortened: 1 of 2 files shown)


core/src/main/java/fj/Equal.java
```diff
@@ -292,6 +292,11 @@ public boolean equal(A a1, A a2) {
   public static final Equal<BitSet> bitSetSequal = equalDef((bs1, bs2) -> bs1.longValue() == bs2.longValue());
 
   /**
+   * An equal instance for the {@link Stream} of {@link Character}.
+   */
+  public static final Equal<Stream<Character>> streamCharacterEqual = streamEqual(charEqual);
```
jbgi: I think this is a bit too much of specific to be part of the public api.
Can we put in a static private nested class of LazyString?

l1cache:
You are right, replacing `LazyString.toStream` with `LazyString::toStream` wouldn't trigger the dead lock. And it also looks a lot cleaner.

Issue #326, opened 2017-10-18
Add ways to turn P2 into P3


janbols:
Hi,

I'm often faced with the situation where I want to make a P2 from other P2's. I end up with something like ``P2<P2<A, B>, C>``. As this becomes unwieldy very quick, I really want to make a P3 instead. 

I don't think there currently is an easy way to do that except by explicitly calling the P3 factory method in the P class like in the following ``P.p(myP2._1(), myP2._2(), c)``. 

However, I think it makes sense to add a method that turns a P2 into a P3 either by calling one of the factory methods in in ``fj.P`` like ``static <A,B,C> P3<A,B,C> p(P2<A,B> fst, C snd)`` 
or by adding an extra method in ``fj.P2`` like ``<C> P3<A,B,C> p(C snd)``.

Would that be a sensible thing to do? Should I make pull requests for these?

Greetings
Jan

jbgi:
:+1: for an extra method in `P2`.
Well, maybe two:
```java
class P2<C> {
  <C> P3<C, A, B> insertFirst(C c);
  <C> P3<A, B, C> insertLast(C c);
}
```
(probably with better names)

Also similar methods would also be added to other `PX` for consistency.

JordanMartinez:
For better names, would it be unwise to use `append` and `prepend`?
````java
class P2<A, B> {
  <C> P3<A, B, C> append(C c);
  <C> P3<C, A, B> prepend(C c);

  // or also
   <C, D> P4<A, B, C, D> append(P2<C, D> p);
   <C, D> P4<C, D, A, B> prepend(P2<C, D> p);
}
`````

**Comment by dobesv** [comment]
It might follow convention to put this into `P`, like:

```java
class P {
  public static P3<A, B, C> flatten(A a, P2<B, C> bc) { ... }
  public static P3<A, B, C> flatten(P2<A, B> ab, C c) { ... }
}
```

janbols:
Ok, thanks for all the suggestions. I added a PR adding ``append`` methods to all Px classes.
I chose for ``append`` as method name because it felt the most explanatory.
I didn't add ``prepend`` methods because I don't know if they are really useful.
I also didn't add factory methods to the ``P`` class because you're always starting from an existing ``Px`` instance so it felt more powerful to add the method on the instance itself.

Please let me know if I need to change anything.

PR #327, opened 2017-10-19
Add ways to turn P2 into P3 #326

janbols:
Added append methods to Px classes for #326

### DIFF
--- a/core/src/main/java/fj/P2.java
+++ b/core/src/main/java/fj/P2.java
@@ -1,12 +1,11 @@
 package fj;
 
-import fj.data.*;
-import fj.data.optic.Lens;
-import fj.data.optic.PLens;
-
 import static fj.Function.*;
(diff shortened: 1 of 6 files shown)


core/src/main/java/fj/P2.java
```diff
@@ -1,12 +1,13 @@
 package fj;
 
-import static fj.Function.*;
```
jbgi: Can you restore the import order as before?

janbols: Ok, just did.

gliptak:
Would we like to see some (basic) tests against this?

jbgi:
@gliptak I think the signature of the methods already enforce a unique possible implementation, so tests are not necessary in this case.

jbgi:
Thanks!

PR #328, opened 2018-01-12
Introduce the Eval monad implementation

izeigerman:
I hope this project is still alive, since it's relevant now more than ever. This PR introduces the `Eval` monad - an abstraction over computation (lazy vs eager). All transformations are lazy and completely stack-safe thanks to the `Trampoline` abstraction. I hope you'll find this useful as well as I do.

### DIFF
--- a/core/src/main/java/fj/data/Eval.java
+++ b/core/src/main/java/fj/data/Eval.java
@@ -1,0 +1,217 @@
+package fj.data;
+
+import fj.F;
+import fj.F0;
+import fj.P;
+import fj.control.Trampoline;
+
+/**
+ * <code>Eval</code> is an abstraction over different models of evaluation.
+ * The data constructors:
+ * <ul>
+ *   <li><code>Now</code> - the value is evaluated immediately.</li>
+ *   <li><code>Later</code> - the value is evaluated only once when it's requested (lazy evaluation).</li>
+ *   <li><code>Always</code> - the value is evaluated every time when it's requested.</li>
+ * </ul>
+ *
+ * Both <code>Later</code> and <code>Always</code> are lazy computations, while <code>Now</code> is eager.
+ *
+ *
+ * @version %build.number%
+ */
+public abstract class Eval<A> {
+
+  /**
+   * Constructs an eager evaluation by wrapping the given value.
+   *
+   * @param a the evaluated value.
+   * @return an eval with computed value.
+   */
+  public static <A> Eval<A> now(A a) {
+    return new Now<>(a);
+  }
+
+  /**
+   * Constructs a lazy evaluation with caching.
+   *
+   * @param a the supplier that evaluates a value.
+   * @return a lazy evaluation.
+   */
+  public static <A> Eval<A> later(F0<A> a) {
+    return new Later<>(a);
+  }
+
+  /**
+   * Constructs a lazy evaluation without caching.
+   *
+   * @param a the supplier that evaluates a value.
+   * @return a lazy evaluation.
+   */
+  public static <A> Eval<A> always(F0<A> a) {
+    return new Always<>(a);
+  }
+
+  /**
+   * Evaluates the computation and return its result.
+   * Depending on whether the current instance is lazy or eager the
+   * computation may or may not happen at this point.
+   *
+   * @return a result of this computation.
+   */
+  public abstract A value();
+
+  /**
+   * Transforms <code>Eval<A></code> into a <code>Eval<B></code> using
+   * the given function.
+   *
+   * Note: the computation of the given transformation is always lazy,
+   * even if it invoked for an eager <code>Now</code> instance. This computation
+   * is performed in O(1) stack space.
+   *
+   * @param f the transformation function.
+   * @return a transformed evaluation.
+   */
+  public <B> Eval<B> map(final F<A, B> f) {
+    return bind(a -> now(f.f(a)));
+  }
+
+  /**
+   * Alias for {@link #bind(F)}.
+   */
+  public <B> Eval<B> flatMap(final F<A, Eval<B>> f) {
+    return bind(f);
+  }
+
+  /**
+   * Transforms <code>Eval<A></code> into a <code>Eval<B></code> using
+   * the given function that directly produces <code>Eval<B></code>.
+   *
+   * Note: the computation of the given transformation is always lazy,
+   * even if it invoked for an eager <code>Now</code> instance. This computation
+   * is performed in O(1) stack space.
+   *
+   * @param f the transformation function.
+   * @return a transformed evaluation.
+   */
+  public <B> Eval<B> bind(final F<A, Eval<B>> f) {
+    if (isTrampoline()) {
+      return bindTrampoline(f, (TrampolineEval<A>) this);
+    } else {
+      return bindTrampoline(f, pureTrampoline(this));
+    }
+  }
+
+  /**
+   * True if the current instance is a trampoline instance, false - otherwise.
+   */
+  boolean isTrampoline() {
+    return false;
+  }
+
+  /**
+   * Represents an eager computation.
+   */
+  private static final class Now<A> extends Eval<A> {
+    private final A a;
+
+    private Now(A a) {
+      this.a = a;
+    }
+
+    @Override
+    public A value() {
+      return a;
+    }
+  }
+
+  /**
+   * Represents a lazy computation that is evaluated only once.
+   */
+  private static final class Later<A> extends Eval<A> {
+    private F0<A> producer;
+    private A a;
+
+    private Later(F0<A> producer) {
+      this.producer = producer;
+    }
+
+    @Override
+    public A value() {
+      if (producer != null) {
+        a = producer.f();
+        producer = null;
+      }
+      return a;
+    }
+  }
+
+  /**
+   * Represents a lazy computation that is evaluated every time when it's requested.
+   */
+  private static final class Always<A> extends Eval<A> {
+    private F0<A> supplier;
+
+    private Always(F0<A> supplier) {
+      this.supplier = supplier;
+    }
+
+    @Override
+    public A value() {
+      return supplier != null ? supplier.f() : null;
+    }
+  }
+
+  /**
+   * A helper abstraction that allows to perform recursive lazy transformations in O(1) stack space.
+   */
+  private static abstract class TrampolineEval<A> extends Eval<A> {
+
+    protected abstract Trampoline<A> trampoline();
+
+    @Override
+    public A value() {
+      return trampoline().run();
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    @Override
+    public A value() {
+      return a;
+    }
+  }
+
+  /**
+   * Represents a lazy computation that is evaluated only once.
```
jbgi: in multi-threaded environment this implementation could result in multiple evaluation.
See https://github.com/functionaljava/functionaljava/blob/v4.7/core/src/main/java/fj/P1.java#L285 for a thread-safe implementation of a lazy thunk.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,217 @@
...
+   *
+   * Note: the computation of the given transformation is always lazy,
+   * even if it invoked for an eager <code>Now</code> instance. This computation
+   * is performed in O(1) stack space.
+   *
+   * @param f the transformation function.
+   * @return a transformed evaluation.
+   */
+  public <B> Eval<B> bind(final F<A, Eval<B>> f) {
+    if (isTrampoline()) {
```
jbgi: instead of an abstract `isTrampoline` I think it would be simpler and safer to have a an abstract `asTrampoline` that return a `TrampolineEval<A>`. Then this method become simply:
`return bindTrampoline(f, this.asTrampoline);`

izeigerman: This is a very good point, thanks.

izeigerman:
@jbgi thanks for reviewing this. The `Eval` is just a nicer/cleaner/easier to use abstraction over the `Trampoline` + `P1`. It has exactly one goal, it can be chained nicely and it hides all the `Trampoline` complexity. It was heavily inspired by the similar implementation in the `Cats` library: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Eval.scala. I found it super useful in Scala and I think Java can benefit from it too.

core/src/main/java/fj/data/Eval.java
izeigerman: Thanks, reused the `P1` memo to solve this.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+  private static final class Always<A> extends Eval<A> {
+    private final F0<A> supplier;
+
+    private Always(F0<A> supplier) {
+      this.supplier = supplier;
+    }
+
+    @Override
+    public A value() {
+      return supplier != null ? supplier.f() : null;
```
jbgi: should just be `return supplier.f()`.
In FJ we do not special on case `null` imput value: protecting each input parameter against null would litter the code with garbage, maybe also impact perf and, in this case, hide root issue by propagating a null value.

Only dedicated APIs like `Option.fromNull` explicitly deal with `null`.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+    return pureTrampoline(this);
+  }
+
+  /**
+   * Represents an eager computation.
+   */
+  private static final class Now<A> extends Eval<A> {
+    private final A a;
+
+    private Now(A a) {
```
jbgi: make the constructor package private instead of private to avoid going through a generated synthetic constructor.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }
+
+  /**
+   * Represents a lazy computation that is evaluated only once.
+   */
+  private static final class Later<A> extends Eval<A> {
+    private final P1<A> memo;
+
+    private Later(F0<A> producer) {
```
jbgi: package private

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }
+
+  /**
+   * Represents a lazy computation that is evaluated every time when it's requested.
+   */
+  private static final class Always<A> extends Eval<A> {
+    private final F0<A> supplier;
+
+    private Always(F0<A> supplier) {
```
jbgi: package private

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+    @Override
+    TrampolineEval<A> asTrampoline() {
+      return this;
+    }
+  }
+
+  private static final class PureTrampolineEval<A> extends TrampolineEval<A> {
+    private final Eval<A> start;
+
+    private PureTrampolineEval(Eval<A> start) {
```
jbgi: package private

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+    protected Trampoline<A> trampoline() {
+      return Trampoline.suspend(P.lazy(() -> Trampoline.pure(start.value())));
+    }
+  }
+
+  private static final class BindTrampolineEval<A, B> extends TrampolineEval<B> {
+    private final TrampolineEval<A> next;
+    private final F<A, Eval<B>> f;
+
+    private BindTrampolineEval(F<A, Eval<B>> f, TrampolineEval<A> next) {
```
jbgi: package private

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+
+  /**
+   * A helper abstraction that allows to perform recursive lazy transformations in O(1) stack space.
+   */
+  private static abstract class TrampolineEval<A> extends Eval<A> {
+
+    protected abstract Trampoline<A> trampoline();
+
+    @Override
+    public A value() {
```
jbgi: could be `final`.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+   * the given function.
+   *
+   * Note: the computation of the given transformation is always lazy,
+   * even if it invoked for an eager <code>Now</code> instance. This computation
+   * is performed in O(1) stack space.
+   *
+   * @param f the transformation function.
+   * @return a transformed evaluation.
+   */
+  public <B> Eval<B> map(final F<A, B> f) {
```
jbgi: could be `final`.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+    return bind(a -> now(f.f(a)));
+  }
+
+  /**
+   * Alias for {@link #bind(F)}.
+   */
+  public <B> Eval<B> flatMap(final F<A, Eval<B>> f) {
```
jbgi: could be `final`.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+   * the given function that directly produces <code>Eval<B></code>.
+   *
+   * Note: the computation of the given transformation is always lazy,
+   * even if it invoked for an eager <code>Now</code> instance. This computation
+   * is performed in O(1) stack space.
+   *
+   * @param f the transformation function.
+   * @return a transformed evaluation.
+   */
+  public <B> Eval<B> bind(final F<A, Eval<B>> f) {
```
jbgi: could be `final`.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+    return bindTrampoline(f, asTrampoline());
+  }
+
+  /**
+   * Transforms the current instance into a trampoline instance.
+   */
+  TrampolineEval<A> asTrampoline() {
```
jbgi: I am not a big fan of override of concrete method. Could you make it abstract? (ie. methods should be either `abstract` or `final`: easier to reason about)

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+
+    protected abstract Trampoline<A> trampoline();
+
+    @Override
+    public A value() {
+      return trampoline().run();
+    }
+
+    @Override
+    TrampolineEval<A> asTrampoline() {
```
jbgi: `final`.

core/src/main/java/fj/data/Eval.java
```diff
@@ -0,0 +1,209 @@
...
+      this.f = f;
+    }
+
+    @Override
+    protected Trampoline<B> trampoline() {
+      return Trampoline.suspend(P.lazy(() -> next.trampoline().map(v -> f.f(v).value())));
+    }
+  }
+
+  private static <A> TrampolineEval<A> pureTrampoline(Eval<A> start) {
```
jbgi: not sure this private methods (`pureTrampoline` and `bindTrampoline`) add much. could you inline them?

izeigerman:
@jbgi thanks for approving this PR. Are there any concerns about merging it?

jbgi:
@izeigerman no concerns, just waiting a bit for people who would like to comment.

@gliptak, @runarorama, @tonymorris any comment? can I merge this?

tonymorris:
LGTM

izeigerman:
While working on a deferred `Eval` computation I've noticed a major flaw in the `BindTrampolineEval` logic: premature evaluation of the nested trampoline instance. As a result recursive computation turned out to be not stack-safe at all. I was going to introduce the `Defer` in the next PR, but since I've discovered this issue, it might as well be introduced as part of existing one. @jbgi and @tonymorris please revisit this update. I'm sorry for the inconvenience, I was sure I tested recursive invocations well enough.

gliptak:
LGTM @izeigerman Thank you for working the tests

izeigerman:
Hey guys, are we good to go here?

core/src/test/java/fj/data/EvalTest.java
```diff
@@ -39,6 +39,20 @@ public void testAlways() {
...
   }
 
+  @Test
+  public void testDefer() {
+    // Make sure that a recursive computation is actually stack-safe.
+    Assert.assertEquals(even(200000).value(), "done");
+  }
+
+  private static Eval<String> even(int n) {
+    return Eval.now(n <= 0).flatMap(b -> b ? Eval.now("done") : Eval.defer(() -> odd(n - 1)));
```
jbgi: from my tests 
```
  private static Eval<String> even(int n) {
    return Eval.now(n <= 0).flatMap(b -> b ? Eval.now("done") : odd(n - 1));
  }

  private static Eval<String> odd(int n) {
    return even(n - 1);
  }
```
works equally well (since you fixed `BindTrampolineEval`) not sure when `Eval.defer` can be useful.

PR #331, opened 2018-02-02
Added some useful functions to Equal

l1cache:


jbgi:
> On a scale 1 to 10, how alive is this project?

I am glad of your optimism shown by not including 0 ;)
But yeah, realistically there is currently few active maintainers so I would say maybe 4 on your scale currently.
I personally try to review PR as quickly as possible, but a recent new-born is taking a lot of my free time.
We are of course welcoming new contributors such as your-self.

A for this current PR, I actually had a look a some times ago, but forgot to get back to you, sorry.
I was pondering if those combinators would be better suited for `F2<A, A, Boolean>` and be in placed in [`Booleans.java`](https://github.com/functionaljava/functionaljava/blob/master/core/src/main/java/fj/function/Booleans.java)?
Or, alternatively, I'd rather have them on the `Definition` interface.

The reason is that I see `Equal` as a type class, and as such it would feel weird to have multiple instances of `Equal<Foo>` in scope as this will be required for using thoses new methods.

how does that sound?

l1cache:
I thought it would be more convenient to have these methods as members, cause it feels natural to call it on Equal.
Also, contramap function basically works the same way.
We can look at methods like this as a functions in the same module as type class. I actually don't see a big difference between this and putting methods like map and filter inside Stream class.
IMHO, type classes and data types are pretty similar in this respect: we define class or data and then we write some functions that use our class or data. In functional languages like haskell we put it in the same module, in OO language we put it in the same class.

And I also like the way it works in java Comparator interface :)

What do you think?

jbgi:
I think I need some example usage of the api, I suspect that if the main usage is as a kind of "builder pattern" for `Equal` we could manage something a bit more straightforward.

l1cache:
"and" and "andThen" are pretty similar, the difference is more philosophical. You would want to use "and" if both Equals are equally important and "andThen" when first Equal is more important and you want to emphasize it.
I couldn't think of any real use cases for "or". I would still keep it for symmetry, but it's ok if you want to remove it.

gliptak:
@jgbi Are you OK with keeping the methods? Thanks

jbgi:
@l1cache @gliptak I pushed slightly different approach on  #333, keeping only the `andThen` APIs, and avoiding intermediate, partially defined, `Equal` or `Ord` instances. Please comment on it.

jbgi:
Closing in favor of #333 (that will be merged soon if no objection)

PR #338, opened 2018-07-21
use openjdk7, remove oracle installer, try sudo false

jbgi:
- Travis does not support oraclejdk7 anymore.
- Oracle-java8-installer may not be necessary (recent enough jdk already provided) and slow down the build.
- Using docker container should also speed up the build.


gliptak:
@jbgi Yes, the change decreases build time by ~40%

PR #343, opened 2018-07-29
Implement Zipper Eq and Hash and add tests

gliptak:

### DIFF
--- a/core/src/main/java/fj/Equal.java
+++ b/core/src/main/java/fj/Equal.java
@@ -1,19 +1,6 @@
 package fj;
 
-import fj.data.Array;
-import fj.data.Either;
-import fj.data.LazyString;
-import fj.data.List;
-import fj.data.Natural;
-import fj.data.NonEmptyList;
-import fj.data.Option;
-import fj.data.Seq;
-import fj.data.Set;
-import fj.data.Stream;
-import fj.data.Tree;
-import fj.data.TreeMap;
-import fj.data.Validation;
-import fj.data.Writer;
+import fj.data.*;
 import fj.data.hamt.BitSet;
 import fj.data.hlist.HList;
 import fj.data.vector.V2;
@@ -453,6 +440,21 @@ public final class Equal<A> {
 
       return x1.isEmpty() && x2.isEmpty();
     });
+  }
+
+  /**
+   * An equal instance for the {@link Zipper} type.
+   *
+   * @param ea Equality across the elements of the zipper.
+   * @return An equal instance for the {@link Zipper} type.
+   */
+  public static <A> Equal<Zipper<A>> zipperEqual(final Equal<A> ea) {
+    Equal<Stream<A>> se = Equal.streamEqual(ea);
+    return equalDef((a1, a2) ->
+        se.eq(a1.lefts(), a2.lefts()) &&
+        ea.eq(a1.focus(), a2.focus()) &&
+        se.eq(a1.rights(), a2.rights())
+    );
   }
 
   /**
--- a/core/src/main/java/fj/Hash.java
+++ b/core/src/main/java/fj/Hash.java
@@ -263,6 +263,26 @@ public final class Hash<A> {
   }
 
   /**
+   * A hash instance for the {@link Zipper} type.
+   *
+   * @param ha A hash for the elements of the zipper.
+   * @return A hash instance for the {@link Zipper} type.
+   */
+  public static <A> Hash<Zipper<A>> zipperHash(final Hash<A> ha) {
+    Hash<Stream<A>> sh = streamHash(ha);
+    return hash(as -> {
+        final int p = 419;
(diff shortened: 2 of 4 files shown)


core/src/main/java/fj/Equal.java
```diff
@@ -455,6 +442,37 @@ public boolean equal(A a1, A a2) {
...
+   * @param ea Equality across the elements of the zipper.
+   * @return An equal instance for the {@link Zipper} type.
+   */
+  public static <A> Equal<Zipper<A>> zipperEqual(final Equal<A> ea) {
+    Definition<A> eaDef = ea.def;
+    return equalDef((a1, a2) -> {
+      Zipper<A> x1 = a1;
+      Zipper<A> x2 = a2;
+
+      x1.move(0);
```
jbgi: those two `move(0)` statements do nothing?

core/src/main/java/fj/Hash.java
```diff
@@ -262,6 +262,32 @@ public int hash(final A a) {
...
+   * @param ha A hash for the elements of the zipper.
+   * @return A hash instance for the {@link Zipper} type.
+   */
+  public static <A> Hash<Zipper<A>> zipperHash(final Hash<A> ha) {
+    return hash(as -> {
+        final int p = 419;
+        int r = 239;
+        Zipper<A> aas = as;
+
+        aas.move(0);
```
jbgi: this does nothing?

core/src/main/java/fj/Equal.java
gliptak: The zipper handed out might not be focus at start. Let me know if you think they are not needed.

core/src/main/java/fj/Hash.java
gliptak: The zipper handed out might not be focus at start. Let me know if you think they are not needed.

core/src/main/java/fj/Equal.java
jbgi: but `move` is a pure function so you must use the result if you want to do that.
But probably the position of a zipper is important and I'd be in favor of just comparing `lefts()`, `focus()` and `right()`. This is also what the haskell implementations I have seen do.

core/src/main/java/fj/Hash.java
jbgi: Same as for `Equal`, I'd say we just need to combine the hash of `lefts()`,` focus()` and `right()`.

core/src/main/java/fj/Equal.java
gliptak: thanks for the pointer. updated

core/src/main/java/fj/Hash.java
gliptak: updated

core/src/main/java/fj/Equal.java
```diff
@@ -455,6 +442,21 @@ public boolean equal(A a1, A a2) {
...
+  /**
+   * An equal instance for the {@link Zipper} type.
+   *
+   * @param ea Equality across the elements of the zipper.
+   * @return An equal instance for the {@link Zipper} type.
+   */
+  public static <A> Equal<Zipper<A>> zipperEqual(final Equal<A> ea) {
+    Definition<A> eaDef = ea.def;
+    return equalDef((a1, a2) -> {
+      return Equal.equals0(Stream.class, a1.lefts(), a2.lefts(), () -> Equal.streamEqual(Equal.anyEqual())) &&
```
jbgi: I think `streamEqual` should use `Equal<A>` not `anyEqual`. also It would be best to save the `streamEqual` in a local variable for reuse.
Also `equals0`need to be used.

core/src/main/java/fj/Hash.java
```diff
@@ -262,6 +262,25 @@ public int hash(final A a) {
...
+   *
+   * @param ha A hash for the elements of the zipper.
+   * @return A hash instance for the {@link Zipper} type.
+   */
+  public static <A> Hash<Zipper<A>> zipperHash(final Hash<A> ha) {
+    return hash(as -> {
+        final int p = 419;
+        int r = 239;
+
+        r = p * r + streamHash(ha).hash(as.lefts());
```
jbgi: also probably best to save `streamHash(ha)` in a local variable.

gliptak: done

core/src/main/java/fj/Equal.java
gliptak: partially updated. I'm not sure about ```equals0```

jbgi:
@gliptak added a commit with remaining observations: https://github.com/functionaljava/functionaljava/pull/343/commits/aa295b9465a1a7c028c10d071981739528031c96
Please squash it/fixup and I will merge.

gliptak:
thank you for the update. squashed.

PR #350, opened 2018-08-05
Implement Vector equals

gliptak:


gliptak:
it was tripping up on coverage ... now it is green

PR #356, opened 2018-08-21
Add DList tests

gliptak:

### DIFF
--- a/core/src/test/java/fj/data/DListTest.java
+++ b/core/src/test/java/fj/data/DListTest.java
@@ -1,0 +1,31 @@
+package fj.data;
+
+import org.junit.Test;
+
+import static fj.data.DList.*;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
+public class DListTest {
+    @Test
+    public void testConsSnoc() {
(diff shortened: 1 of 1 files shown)


core/src/test/java/fj/data/DListTest.java
```diff
@@ -0,0 +1,31 @@
...
+public class DListTest {
+    @Test
+    public void testConsSnoc() {
+        assertThat(nil().snoc(2).cons(1).toJavaList(), is(single(1).snoc(2).toJavaList()));
+    }
+
+    @Test
+    public void testListDList() {
+        DList<Integer> d = listDList(List.range(0, 1000));
+        assertThat(d.toJavaList(), is(d.run().toJavaList()));
```
jbgi: instead of testing `assertThat(d.toJavaList(), is(d.run().toJavaList()));` (that only check that `toJavaList()` is an alias of `run().toJavaList()`) I think we better test something like 
`assertThat(d.toJavaList(), is(List.range(0, 1000).toJavaList()));`
(same for below methods)

gliptak: updated

Issue #358, opened 2018-08-23
Fix code so compiles with JDK 10


gliptak:
https://github.com/functionaljava/functionaljava/issues/355

gliptak:
https://github.com/functionaljava/functionaljava/pull/359

jbgi:
Fixed by ebdceaa7653780d8c896ad08687a22b934538aa5

PR #359, opened 2018-08-25
Target Java 10, fix compile, move java8 to core. Fix #355, #358

### LINKED ISSUE #355: Add current JDK to Travis build (allowed failure?)
https://www.deps.co/guides/travis-ci-latest-java/

jbgi:


jbgi:
so neither jdk10, 11 or 12 are able to build the project for now. Weird jdk bugs...

core/src/main/java/fj/data/TreeZipper.java
```diff
@@ -627,6 +627,9 @@ else for (final TreeZipper<A> loc : parent())
             }, f.f(this));
   }
 
+  /**
+   * @param <A>
```
gliptak: This might be triggering the Javadoc fail
```
javadoc: error - An internal exception has occurred. 
	(com.sun.tools.javac.code.ClassFinder$BadClassFile: bad class file: /home/travis/build/functionaljava/functionaljava/core/build/libs/functionaljava-5.0-SNAPSHOT.jar(/fj/data/TreeZipper$2.class)
  undeclared type variable: A
```

jbgi: ah I added that in hope to fix the issue. But that did not solve it. Will remove.

jbgi:
Solved [jdk10 bug](https://bugs.openjdk.java.net/browse/JDK-8187422) by refactoring some anonymous classes as private static classes.
Made JDK11+ optional for now. Did not investigate the cryptic error.

gliptak:
Should we consider submitting a JDK11 bug report?

jbgi:
> Should we consider submitting a JDK11 bug report?

Yes but we should try to minimize the issue first. Appear to come from a class of the quickcheck module. Probably one of the classes responsible from launching the tests, dealing with threads and annotations.

jbgi:
I workaround the last issue under JDK11+. Opened a jdk bug with a "minimal" test case (https://github.com/jbgi/jdk11-big-integer-substract-bug).

The question is, should we merge this in master or in a `5.x` branch?

gliptak:
would ```master``` targeting a 5.0 release (soon) work?

jbgi:
I'd expect the cleanup necessary for a 5.0 release to take a few months, and, at the very least, we have to wait for jdk11 release as it would make sense to target it as it's going to be an LTS version (people are going to switch directly from jdk8 to jdk11).
So it's probably best to create a `series/5.x` branch for now. I will do some cleanup and then (if jdk 11 is released) we would switch the default branch to  `series/5.x`.
To be consistent we should probably also rename current master to  `series/4.x`.
how does that sounds?

gliptak:
I didn't realize that there is still a significant cleanup effort. And good point on the expected JDK transition. Above approach sounds good.

jbgi:
> I didn't realize that there is still a significant cleanup effort. 

Maybe not, but I'd like to also take the opportunity to explore new ideas, like deriving some stuff via derive4j.

@gliptak also could you give me owner rights, so that I can setup default branches correctly?

gliptak:
@jbgi changed your role to Owner

Issue #371, opened 2018-10-08
Investiguate/fix failing property test: BitSetProperties.rangeTest


jbgi:
This test fails for input
```java
x = 8205295599903022080L;
list = List.list(19,19,19,21,30,53,57);
```

petervasil:
Hello, can I help with this issue?

jbgi:
@petervasil that would be awesome, yes!

SoundharyaKamaraj:
Can I work on it?

gliptak:
@petervasil  @SoundharyaKamaraj  patches are welcome

PR #373, opened 2018-10-20
#371: fixed failing property test 

SoundharyaKamaraj:


jbgi (approved):
I see, so hopefully the bug was not in the implementation but in the test.
Thanks for the investigation and fix!

Issue #363, opened 2018-10-02
Implement separate set union and set intersection monoids (and semigroups)


MarekDudek:
Currently monoid for set is union monoid, we could use both.

jbgi:
:+1: for the intersection Semigroup.

For the intersection `Monoid<Set<A>>`, the zero would need to be the set of all values for that type , which would requires instance of `Enumerator<A>` (we have) and `Bounded<A>` (we don't have yet).

So this can be separated into two PRs if more manageable.

toxyduck:
Hello, I would like to contribute. Can you explain, Why do I need to use `Bounded<A>` for intersection monoid zero, if `Enumerator<A>` can provide max and min elements?

jbgi:
max and min in Enumerator only return Option values and thus this is not very practical to instanciate the intersection Monoid. We could pass min and max explicitly though and not create Bounded for now.

PR #391, opened 2019-10-20
Implement intersection monoid and semigroup

toxyduck:
#363

### DIFF
--- a/core/src/main/java/fj/Bounded.java
+++ b/core/src/main/java/fj/Bounded.java
@@ -2,9 +2,41 @@ package fj;
 
 public class Bounded<A> {
 
+    private final Definition<A> def;
+
     public interface Definition<A> {
         A min();
 
         A max();
+    }
+
+    private Bounded(Definition<A> definition) {
+        this.def = definition;
+    }
+
+    public A min() {
+        return def.min();
+    }
+
+    public A max() {
+        return def.max();
+    }
+
+    public static <A> Bounded<A> boundedDef(Definition<A> def) {
+        return new Bounded<>(def);
+    }
+
+    public static <A> Bounded<A> bounded(A min, A max) {
+        return boundedDef(new Definition<A>() {
+            @Override
+            public A min() {
+                return min;
+            }
+
+            @Override
+            public A max() {
+                return max;
+            }
+        });
     }
 }
--- a/core/src/main/java/fj/Bounded.java
+++ b/core/src/main/java/fj/Bounded.java
@@ -1,0 +1,10 @@
+package fj;
+
+public class Bounded<A> {
+
+    public interface Definition<A> {
+        A min();
+
+        A max();
+    }
+}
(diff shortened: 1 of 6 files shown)


toxyduck:
@jbgi Please, review

core/src/main/java/fj/Bounded.java
```diff
@@ -0,0 +1,50 @@
+package fj;
+
+/**
+ * The Bounded class is used to name the upper and lower limits of a type.
+ * Ord is not a superclass of Bounded since types that are not totally ordered may also have upper and lower bounds.
+ */
+public class Bounded<A> {
```
jbgi:
```suggestion
public final class Bounded<A> {
```

core/src/main/java/fj/data/Enumerator.java
```diff
@@ -185,6 +181,18 @@ private Enumerator(final F<A, Option<A>> successor, final F<A, Option<A>> predec
...
 
+  /**
+   * Returns a stream of the values from this enumerator,
+   * starting at the min of given Bounded, ending at the max, counting up.
+   *
+   * @param bounded A value at which to begin the stream.
+   * @return a stream of the values from this enumerator, cut by bounded, counting up.
+   */
+  public Stream<A> toStream(final Bounded<A> bounded) {
+    final F<A, A> id = identity();
```
jbgi: I think the implementation of this method should just delegate to the existing overload:
```
public Stream<A> toStream(final Bounded<A> bounded) {
  return toStream(bounded.min());
}
```

core/src/main/java/fj/Monoid.java
```diff
@@ -1109,6 +1102,28 @@ public Unit append(Unit a1, Unit a2) {
...
 
+  /**
+   * A intersection monoid for sets.
+   *
+   * @param bounded A bound for all possible elements
+   * @param enumerator An enumerator for all possible elements
+   * @param o An order for set elements.
+   * @return A monoid for sets whose elements have the given order.
+   */
+  public static <A> Monoid<Set<A>> setIntersectionMonoid(final Bounded<A> bounded, final Enumerator<A> enumerator, final Ord<A> o) {
```
jbgi: here you can remove the `final Ord<A>` parameter since it is available through `enumerator.order()`.

jbgi (changes requested):
Great PR, thanks! some small adjustments and we can merge.

core/src/main/java/fj/data/Enumerator.java
```diff
@@ -185,6 +181,18 @@ private Enumerator(final F<A, Option<A>> successor, final F<A, Option<A>> predec
...
+    return Stream.fromFunction(this, id, bounded.min()).takeWhile(item -> !item.equals(bounded.max()));
```
jbgi: takeWhile should be 
```suggestion
    return Stream.fromFunction(this, id, bounded.min()).takeWhile(item ->order.isLessThanOrEqualTo(item, bounded.max()));
```
otherwise you miss Bounded.max() in the Stream. (also we avoid the use `Object#equals`).

jbgi (approved):

PR #392, opened 2019-10-28
Basic JPMS (modules) support through 'Automatic-Module-Name'

gneuvill:
Currently, having the 'functionaljava_1.8' jar on the classpath of a jvm version 9 or above is problematic since the module name automatically derived by either javac or the vm is broken. This P.R fixes this.

Whole modules declaration (module-info.java) should happen on the 5.x branch, I think.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -181,6 +181,7 @@ configure(subprojects.findAll { it.name != "props-core" }) {
         buildCommand 'org.eclipse.pde.ManifestBuilder'
         buildCommand 'org.eclipse.pde.SchemaBuilder'
       }
+      instruction 'Automatic-Module-Name', "functionaljava${project.name == 'core' ? '' : "-$project.name"}"
     }
 
     // Output MANIFEST.MF statically so eclipse can see it for plugin development
--- a/build.gradle
+++ b/build.gradle
@@ -172,7 +172,7 @@ configure(subprojects.findAll { it.name != "props-core" }) {
             if(project.name != "core") {
                 instruction 'Require-Bundle', 'org.functionaljava;bundle-version="'+project.fjBaseVersion+'"'
             }
-            instruction 'Automatic-Module-Name', "functionaljava${project.name == 'core' ? '' : "-$project.name"}"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -169,9 +169,10 @@ configure(subprojects.findAll { it.name != "props-core" }) {
             instruction 'Signature-Version', project.fjVersion
             instruction 'Bundle-ActivationPolicy', 'lazy'
             instruction 'Bundle-Vendor', 'functionaljava.org'
-			if(project.name != "core") {
-				instruction 'Require-Bundle', 'org.functionaljava;bundle-version="'+project.fjBaseVersion+'"'
-			}
+            if(project.name != "core") {
+                instruction 'Require-Bundle', 'org.functionaljava;bundle-version="'+project.fjBaseVersion+'"'
+            }
+            instruction 'Automatic-Module-Name', "functionaljava${project.name == 'core' ? '' : ".$project.name"}"
```
gliptak, line 175: Only need to apply when ```core````?

gliptak:
@gneuvill please update the PR with the error you are seeing

build.gradle
gneuvill: No, an `Automatic-Module-Name` attribute will be inserted in the manifests of each of the sub-projects with the value :

- for core : 'functionaljava'
- for quickcheck : 'functionaljava.quickcheck'
- etc...

gneuvill: No, an `Automatic-Module-Name` attribute will be inserted in the manifests of each of the sub-projects with the value :

- for core : 'functionaljava'
- for quickcheck : 'functionaljava.quickcheck'
- etc...

jbgi:
@gneuvill  Could we use this instead: https://github.com/java9-modularity/gradle-modules-plugin#separate-compilation-of-module-infojava
?

gliptak:
> @gneuvill Could we use this instead: https://github.com/java9-modularity/gradle-modules-plugin#separate-compilation-of-module-infojava
> ?

For above page shows "This plugin requires JDK 11 to be used when running Gradle." so Travis build might need to be reorganized

jbgi (approved):

PR #400, opened 2020-05-20
Add fj.data.Option.sequence(Validation<E, Option<A>>).

drewctaylor:
Implementation uses traverseOption:

```java
public static <E, A> Option<Validation<E, A>> sequence(final Validation<E, Option<A>> a) {
    return a.traverseOption(identity());
}
```


tonymorris (approved):
Thanks again!

PR #440, opened 2023-04-25
Refactor Junit4 to Junit 5

JLLeitschuh:
Refactors all tests to use Junit 5 instead of Junit 4.

Also brings all files changed into sync with the prevailing format of the codebase which seems to be a two space indent.

Use this link to re-run the recipe: https://public.moderne.io/recipes/builder/nEvzN


JLLeitschuh:
This also needs updates to the `build.gradle` files, but I'll have to do that manually. I want to make sure this is a change that the maintainers would actually appreciate before I move forward with that.

